PR #77, opened 2024-07-08
Feature/add metadata contact#5659

HavierD:
add metadata contact, with some upgrade and bug fixing. 

image

### DIFF
--- a/src/pages/detail-page/subpages/components/NavigatablePanel.tsx
+++ b/src/pages/detail-page/subpages/components/NavigatablePanel.tsx
@@ -20,6 +20,7 @@ const NavigatablePanel: React.FC<NavigatablePanelProps> = ({
 }) => {
   const bigPosition = 99999;
   const panelVisibleHeight = 1480;
+  const resizeDelay = 400;
   const [scrollDistance, setScrollDistance] = useState<number | null>(null);
   const scrollableSectionRef = useRef<HTMLDivElement>(null);
   const firstRef = useRef<HTMLDivElement>(null);
@@ -41,6 +42,13 @@ const NavigatablePanel: React.FC<NavigatablePanelProps> = ({
     });
     setScrollDistance(null);
   }, [scrollDistance]);
+
+  // For better scrolling animation, resizing happens after scrolling
+  const lateResize = useCallback((toReSize: number) => {
+    setTimeout(() => {
+      setSupplimentaryHeight((prevHeight) => prevHeight + toReSize);
+    }, resizeDelay);
+  }, []);
 
   const getRefBy = useCallback((index: number) => {
     switch (index) {
@@ -127,7 +135,11 @@ const NavigatablePanel: React.FC<NavigatablePanelProps> = ({
       const visibleHeight = scrollableSectionRef.current.clientHeight;
       const neededHeight =
         targetPosition - (currentScrollHeight - visibleHeight);
-      setSupplimentaryHeight((prevHeight) => prevHeight + neededHeight);
+      if (neededHeight >= 0) {
+        setSupplimentaryHeight((prevHeight) => prevHeight + neededHeight);
+      } else {
+        lateResize(neededHeight);
+      }
       setScrollDistance(targetPosition);
     };
   };
--- a/src/pages/detail-page/subpages/components/NavigatablePanel.tsx
+++ b/src/pages/detail-page/subpages/components/NavigatablePanel.tsx
@@ -114,6 +114,24 @@ const NavigatablePanel: React.FC<NavigatablePanelProps> = ({
     []
   );
 
+  const onNavigate = (index: number) => {
+    return () => {
+      const ref = getRefBy(index);
+      if (!scrollableSectionRef.current || !ref?.current) return;
+      const targetPosition = ref.current.offsetTop;
+
+      // Calculate the needed height to scroll to the target position
+      const currentScrollHeight = basePointRef.current
+        ? basePointRef.current.offsetTop
+        : 0;
+      const visibleHeight = scrollableSectionRef.current.clientHeight;
+      const neededHeight =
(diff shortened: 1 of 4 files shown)


src/pages/detail-page/subpages/components/NavigatablePanel.tsx
```diff
@@ -13,13 +19,37 @@ const NavigatablePanel: React.FC<NavigatablePanelProps> = ({
   isLoading,
 }) => {
   const bigPosition = 99999;
+  const panelVisibleHeight = 1480;
+  const resizeDelay = 400;
```
vietnguyengit, line 23: fixed arbitrary constants should be clearly distinct from others

```javascript
const BIG_POSITION = 99999;
const PANEL_VISIBLE_HEIGHT = 1480;
const RESIZE_DELAY = 400;
```
also move them out to the top, but below the imports. also add some comments for why those numbers but not something else, or a simple explanation.

src/pages/detail-page/subpages/components/NavigatablePanel.tsx
```diff
@@ -13,13 +19,37 @@ const NavigatablePanel: React.FC<NavigatablePanelProps> = ({
...
+  const [scrollDistance, setScrollDistance] = useState<number | null>(null);
   const scrollableSectionRef = useRef<HTMLDivElement>(null);
   const firstRef = useRef<HTMLDivElement>(null);
   const secondRef = useRef<HTMLDivElement>(null);
   const thirdRef = useRef<HTMLDivElement>(null);
   const fourthRef = useRef<HTMLDivElement>(null);
+  const basePointRef = useRef<HTMLDivElement>(null);
```
vietnguyengit, line 30: should these have `| null` too? e.g 

```javascript
const scrollableSectionRef = useRef<HTMLDivElement | null>(null);
```

src/pages/detail-page/subpages/components/NavigatablePanel.tsx
```diff
@@ -133,10 +177,9 @@ const NavigatablePanel: React.FC<NavigatablePanelProps> = ({
...
       <Grid
         item
-        container
         md={9}
         ref={scrollableSectionRef}
-        sx={{ height: "500px", overflowY: "auto" }}
+        sx={{ height: panelVisibleHeight + "px", overflowY: "auto" }}
         onScroll={(event) => {
           const scrollPosition = event.currentTarget.scrollTop;
           debouncee(scrollPosition);
```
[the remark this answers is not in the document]
vietnguyengit: consider renaming this deliberately incorrect spelling, make it clearer, eg. debouncedScrollHandler  

also I noticed that the current block to initialise `debouncee` appears to be initialised on every new renders unnecessarily.  Initialising it inside useEffect to ensure it runs only once.

you should also have a strategy to cancel any debounce timeouts to avoid memory leaks

https://stackoverflow.com/questions/29012922/how-to-cancel-a-debounced-function-after-it-is-called-and-before-it-executes

so changes may look like

```javascript
  const debouncedScrollHandler = useRef<((number: number) => void) | null>(null);

  useEffect(() => {
    debouncedScrollHandler.current = _.debounce((scrollPosition: number) => {
      setPosition(scrollPosition);
    }, 300);

    return () => {
      debouncedScrollHandler.current?.cancel();
    };
  }, []);

  const handleScroll = (event) => {
    if (debouncedScrollHandler.current) {
      debouncedScrollHandler.current(event.currentTarget.scrollTop);
    }
  };

// and then in component's onScroll, you can use onScroll={handleScroll}
```

src/pages/detail-page/subpages/tab-panels/MetadataInformationPanel.tsx
```diff
@@ -1,7 +1,48 @@
...
+import { useDetailPageContext } from "../../context/detail-page-context";
+import NavigatablePanel from "../components/NavigatablePanel";
+import ContactBlock from "../components/ContactBlock";
 
 const MetadataInformationPanel = () => {
-  return <div>MetadataInformationCard</div>;
+  const context = useDetailPageContext();
+  const metadataContact = context.collection
+    ?.getContacts()
+    ?.filter((contact) => contact.roles.includes("metadata"));
```
vietnguyengit, line 10: not a too big deal, but consider using useMemo

```javascript
const metadataContacts = useMemo(() => {
  return context.collection?.getContacts()?.filter((contact) => contact.roles.includes("metadata")) || [];
}, [context.collection]);
```

src/pages/detail-page/subpages/tab-panels/MetadataInformationPanel.tsx
```diff
@@ -1,7 +1,48 @@
...
+
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    if (!metadataContact) {
+      setIsLoading(true);
+    } else {
+      setIsLoading(false);
+    }
+  }, [metadataContact]);
```
vietnguyengit, line 20: if you use suggested change for metadataContacts, this can become:

```javascript
useEffect(() => {
  setIsLoading(metadataContacts.length === 0);
}, [metadataContacts]);
```

which prevents toggling between true and false unnecessarily, causing fewer re-renders.

vietnguyengit (changes requested):

src/pages/detail-page/subpages/tab-panels/MetadataInformationPanel.tsx
vietnguyengit: for performance

HavierD: Thank you for reminding. It should use useMemo. I forgot it.

vietnguyengit (approved):
LGTM

PR #89, opened 2024-07-16
Feature/5733 map click function

NekoLyn:
1. added click function for removing spatial extents when click on empty space
2. created fixed-on-top area that displays selected datasets cards


utas-raymondng (approved):
LGTM

PR #91, opened 2024-07-17
Feature/add citation and constraints#5704

HavierD:


src/types/DataStructureTypes.tsx
```diff
@@ -31,6 +31,12 @@ interface ITemporal {
   revision?: string;
 }
 
+interface ICitation {
```
utas-raymondng, line 34: Can we move all these back to the store, we should have defined a OGCCollection with variable and possible access. Aim is to remove the DataStructureTypes.tsx

utas-raymondng (changes requested):
As per comment

PR #95, opened 2024-07-22
:bug: fix search bar category bug and enter event bug

NekoLyn:
1. fix search bar filter category bug. This bug can be tested in search result page:
select some categories from filter -> remove them from search input bar -> resize the window to change map bbox to trigger fetching datasets automatically -> go to network see if query param string is correct or not

2. fix search bar enter event bug. This bug can be tested in landing page:
enter 'imos' in search input bar quickly and press enter -> can do search successfully and nav to search result page

### DIFF
--- a/src/components/search/InputWithSuggester.tsx
+++ b/src/components/search/InputWithSuggester.tsx
@@ -240,10 +240,13 @@ const InputWithSuggester: FC<InputWithSuggesterProps> = ({
   const handleKeyDown = (
     event: React.KeyboardEvent<HTMLTextAreaElement | HTMLInputElement>
   ) => {
-    if (event.key === "Enter") {
+    if (event.key === "Enter" && open === true) {
       setOpen(false);
+      handleEnterPressed(event, false);
     }
-    handleEnterPressed(event, open);
+    if (event.key === "Enter" && open === false) {
+      handleEnterPressed(event, false);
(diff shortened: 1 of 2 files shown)


utas-raymondng (approved):
LGTM

src/components/search/InputWithSuggester.tsx
HavierD, line 1: All the implementation is good! Just one minor change request. Could you optimize this part of changes? actually it can be simplified like this:
```javascript

  const handleKeyDown = (
    event: React.KeyboardEvent<HTMLTextAreaElement | HTMLInputElement>
  ) => {
    if (event.key === "Enter") {
      setOpen(false);
    }
    handleEnterPressed(event, false);
  };

```

only the second parameter of the `handleEnterPressed()` needs to change.

NekoLyn:
@HavierD  Thank you Havier! That looks more clean. There is some space for improvement. But I would prefer this way:
`const handleKeyDown = ( event: React.KeyboardEvent<HTMLTextAreaElement | HTMLInputElement> ) => { if (event.key === "Enter") { if (open) { setOpen(false); } handleEnterPressed(event, false); } };`
therefore setOpen(false) and handleEnterPressed(event, false) is called conditionally to avoid redundant state updates and match user behaviour.

PR #97, opened 2024-07-22
:bug: fix grid view

NekoLyn:


src/components/result/ResultCards.tsx
```diff
@@ -47,29 +47,86 @@ const renderCells = (
...
+  /**
+   *if only one dataset is selected:
+   * the first row contains a invisible card related to the selected dataset, and also the first one of result datasets
+   * the following rows are the rest of result datasets
+   * but leftIndex and rightIndex need to minus 1 because one place is taken by the selected dataset card
+   *
+   * if more than one datasets are selected:
+   * then use the leftIndex and rightIndex as normal
+   */
+  if (props.datasetsSelected && props.datasetsSelected?.length === 1) {
```
utas-raymondng, line 59: if you implemented the spider, there will always be 1 item selected,  do you need this check?

utas-raymondng (changes requested):
Feedback needed

src/components/result/ResultCards.tsx
NekoLyn: It means if more than one datasets are selected (for now we still can select multiple dataset in one click), the list will be separate into two parts: upper section is fixed on top that render multiple selected datasets; lower section is the result cards which use the leftIndex and rightIndex as normal.

NekoLyn: if the spider diagram is implemented, there will be only one dataset can be selected in one click. Besides @quye04 may change the UI of showing selected result card. So I block this ticket and wait for design and spider diagram implementation [#5751](https://github.com/aodn/backlog/issues/5751)

PR #98, opened 2024-07-22
refactor remove onClickPopup and add static layers

utas-raymondng:
@NekoLyn can you help me to check if you still need the onClickPopup? Looks like not required


NekoLyn:
@utas-raymondng  we don't need click popup for now, please delete it, thanks!

NekoLyn:
it looks good to me! 
just a small concern that we may need a rule to decide which layer should stay on top which on bottom?
image

utas-raymondng:
The layer always at bottom I think

NekoLyn (approved):

PR #100, opened 2024-07-23
Feature/add license#5704

HavierD:
also includes some refactorings


src/pages/detail-page/subpages/tab-panels/CitationPanel.tsx
```diff
@@ -70,7 +77,11 @@ const CitationPanel = () => {
       },
       {
         title: "License",
-        component: <div>License</div>,
+        component: license ? (
+          <LicenseBlock license={license} />
+        ) : (
+          <div>no license</div>
+        ),
```
NekoLyn, line 84: do we have any design/plan if the dataset has no particular section?

NekoLyn: I think it would be better to not render that part if no data

NekoLyn (approved):

src/pages/detail-page/subpages/tab-panels/CitationPanel.tsx
HavierD: Thank you. this consideration exists for all the part of the details. I just raised a ticket for it to implement empty block https://github.com/aodn/backlog/issues/5780            Will do it in the future

PR #109, opened 2024-08-01
Feature/5751 add spider diagram to map

NekoLyn:


src/components/map/mapbox/component/SpiderDiagram.tsx
```diff
@@ -0,0 +1,410 @@
...
+      const numPins = datasets.length;
+
+      // Spider configuration
+      const circleSpiralSwitchover = 9;
+      const circleFootSeparation = 25;
+      const spiralFootSeparation = 28;
+      const spiralLengthStart = 15;
+      const spiralLengthFactor = 4;
+
+      const generateSpiderLegParams = (count: number) => {
```
utas-raymondng, line 136: Move this out of this function as it is pretty much fixed

utas-raymondng (changes requested):
Minor comment

utas-raymondng:
Few more comments but can fixed in next iteration, please raise tickets to capture these and other comments after the end of iteration review

1. I think we can reduce the zoom level a bit more so we can see the spider earlier
2. Once the spider appears, if you zoom out then the spider still there but keep reducing, that is spider do not disappear
3. Very good the color change on end node thanks.

I am ok to approve the PR once the comment inside the code addressed

src/components/map/mapbox/component/SpiderDiagram.tsx
```diff
@@ -0,0 +1,410 @@
...
+
+interface SpiderDiagramProps extends LayersProps {
+  spiderDiagramConfig?: Partial<SpiderDiagramConfig>;
+  clusterLayer: string;
+  clusterSourceId: string;
+  unclusterPointLayer: string;
+}
+
+const defaultSpiderDiagramConfig: SpiderDiagramConfig = {
+  spiderifyFromZoomLevel: 11,
```
utas-raymondng, line 23: May be try change to 8 instead of 11

utas-raymondng (changes requested):
Once more minor comment

NekoLyn:
> Few more comments but can fixed in next iteration, please raise tickets to capture these and other comments after the end of iteration review
> 
> 1. I think we can reduce the zoom level a bit more so we can see the spider earlier
> 2. Once the spider appears, if you zoom out then the spider still there but keep reducing, that is spider do not disappear
> 3. Very good the color change on end node thanks.
> 
> I am ok to approve the PR once the comment inside the code addressed

Thanks Raymond! I agree to spiderify earlier, but it may take some time to adjust the function of calculating the radius of the spider diagram and legs/pins size to scale up/down the diagram when changing the spiderifyFromZoomLevel.

For the second point I will fix it by making the unspiderify more sensitive to the zoom out (for now it will disappear with zoom out but not so quickly) So I will raise another ticket to fix it together with other function/style issues of spider diagram in next iteration.

utas-raymondng (approved):
LGTM

utas-raymondng:
Please raise ticket for remain change and put links here thanks

NekoLyn:
> Please raise ticket for remain change and put links here thanks

 A ticket to adjust the variable spiderifyFromZoomLevel to make the spiderify appear more earlier https://github.com/aodn/backlog/issues/5810

PR #116, opened 2024-08-07
Feature/5743 improve searchbar

NekoLyn:


src/components/search/InputWithSuggester.tsx
```diff
@@ -135,50 +138,90 @@ const InputWithSuggester: FC<InputWithSuggesterProps> = ({
...
+              (suggestion) => {
+                if (commonSuggestions.includes(suggestion)) {
+                  return { text: suggestion, group: OptionGroup.COMMON };
+                } else if (categorySuggestions.has(suggestion)) {
+                  return { text: suggestion, group: OptionGroup.CATEGORY };
+                } else {
+                  return { text: suggestion, group: OptionGroup.PHRASE };
+                }
+              }
+            );
```
vietnguyengit, line 209: all good but for readability, you might want to extract generic functions from this block, they extracted methods stay outside the `.then()`, and **maybe** reusable by different components in future. e.g.

```javascript
const sortStringList = (stringList: string[], inputValue: string) => {
...
}
```

then call it (line 187)

```javascript
const sortedSuggestions = sortStringList([...allSuggestions], inputValue);
```

src/components/search/InputWithSuggester.tsx
```diff
@@ -135,50 +138,90 @@ const InputWithSuggester: FC<InputWithSuggesterProps> = ({
...
+            }
+
+            // Create array of all unique suggestions
+            const allSuggestions = [
+              ...new Set([
+                ...commonSuggestions,
+                ...categorySuggestions,
+                ...phrasesSuggestions,
+              ]),
+            ];
```
vietnguyengit, line 184: if you extracting method, you don't need to combined actions here, it can be just a `Set`, and leave the converting task for different object.

```javascript
  // Create array of all unique suggestions
  const allSuggestions = new Set([
    ...commonSuggestions,
    ...categorySuggestions,
    ...phrasesSuggestions
  ]);
```

when you pass it to the extracted method, pass as `[...allSuggestions]`, It's OK, as it works and behaves exactly the same whether breaking it down or not. I'd try to avoid combining responsibilities/actions by a single object, that will be helpful when debugging.

src/components/search/InputWithSuggester.tsx
NekoLyn: Thanks Viet! I have extracted a util function and move to utils folder

src/components/search/InputWithSuggester.tsx
NekoLyn: Yes I leave it here the reduce the duplication, but not in the extracted function

vietnguyengit: you meant no changes or you will convert on the fly when passing to the method?

e.g

```javascript
const sortedSuggestions = sortByRelevance([...allSuggestions], inputValue);
```

vietnguyengit: you are current using

```javascript
const sortedSuggestions = sortByRelevance(allSuggestions, inputValue);
```

because you already converted `sortedSuggestions` to `list` type while generating values to it as a `Set`.

src/utils/sortByRelevance.tsx
```diff
@@ -0,0 +1,21 @@
+import { sortBy } from "lodash";
+
+/**
+ * Sorts an array of strings based on their relevance to a target string.
+ *
+ * @param items An array of strings to sort
+ * @param target The target string to compare against
+ * @returns A new array of sorted strings
+ */
+export function sortByRelevance(items: string[], target: string): string[] {
```
vietnguyengit, line 10: as you've created `utils`, I'd suggest to have a generic component which can be grown later. So instead of  `sortByRelevance.tsx` as standalone component, which when you import the method, it looks like

```javascript
import { sortByRelevance } from "../../utils/sortByRelevance";
import { newUtilMethod } from "../../utils/newUtilMethod";
import { anotherUtilMethod } from "../../utils/anotherUtilMethod";
.
.
.
import { oneMoreUtilMethod } from "../../utils/oneMoreUtilMethod";
```

please create something like `helpers.tsx` or `commonUtils.tsx` or whatever, normally, it'd be called `helpers`. And when you import such file, it looks like this

```javascript
import { sortByRelevance, newUtilMethod, anotherUtilMethod, ...   } from "../../utils/helper";
```

src/components/search/InputWithSuggester.tsx
NekoLyn: since I have spread the Set here 
`            const allSuggestions = [
              ...new Set([
                ...commonSuggestions,
                ...categorySuggestions,
                ...phrasesSuggestions,
              ]),
            ];
`
therefore I pass `allSuggestions` in `sortByRelevance` directly.
what you mentioned is a good point: keep allSuggestions as a Set, and modify the extracted function like:
`function sortByRelevance(items: string[] | Set<string>, target: string): string[] {
  const itemsArray = Array.from(items);
  // ... rest of the sorting logic
}
`

src/utils/sortByRelevance.tsx
NekoLyn: That is another good way to organise the utils. Basically now I just fowllow current the file structure , I found there is already a utils folder for different utils functions.

src/utils/sortByRelevance.tsx
```diff
@@ -7,8 +7,12 @@ import { sortBy } from "lodash";
  * @param target The target string to compare against
  * @returns A new array of sorted strings
  */
-export function sortByRelevance(items: string[], target: string): string[] {
-  return sortBy(items, [
+export function sortByRelevance(
```
vietnguyengit, line 10: please also rename this file and make it a common places for future helper methods, **OR** move this method to the existing helpers file if there is one already. This kind of method shouldn't be a standalone component.

vietnguyengit (changes requested):

src/utils/sortByRelevance.tsx
vietnguyengit: right now, it is only 1 - seems no problems at all, but you will have a mess in future if keep creating new standalone components this way unnecessarily, hard to read, hard to maintain, hard to debug.

NekoLyn: Thanks Viets. I have refactored most of the utils folder, but didn't move files like AppRouter.tsx because for now hard to find a place to put this. I will do this in other ticket for updating project folder/file structure.

vietnguyengit: there are `TODO` added to them, should be enough for this scope.

vietnguyengit (approved):

PR #119, opened 2024-08-14
Feature/5649 add apply button in filter

NekoLyn:
1. Add 'apply', 'clear all' and 'X' button in filter
2. refactor DateRangeFilter 
3. move styled components to /components/common
4. add comment 'Unused components' on the top if the code is not in use for now (can be removed later if not use any more)
5. removed unused files in 4


src/components/common/filters/DateRangeFilter.tsx
```diff
@@ -0,0 +1,286 @@
...
+import TimeRangeBarChart from "../charts/TimeRangeBarChart";
+import { OGCCollections } from "../store/OGCCollectionDefinitions";
+import { useDispatch } from "react-redux";
+import { AppDispatch } from "../store/store";
+import { fetchResultNoStore } from "../store/searchReducer";
+import { cqlDefaultFilters } from "../cqlFilters";
+import PlainDatePicker from "../datetime/PlainDatePicker";
+import PlainSlider from "../slider/PlainSlider";
+
+const datePikerSlotProps = {
```
HavierD, line 31: This file is confusing for me. Starting from here there are 35 duplicated codes; The component in this file is not consistent with the file name (`DateRangeSlider` vs `DateRangeFilter`); the name `DateRangeSlider` is also the same with another component. Please double-check this file

src/components/common/store/componentParamReducer.tsx
```diff
@@ -213,14 +212,19 @@ const flattenToProperties = (
...
 // Change the flatten json to xxx=yyy?xxxx=yyyy which can be use in url.
 const formatToUrlParam = (param: ParameterState) => {
   const result = flattenToProperties(param);
   const parts = [];
   for (const key in result) {
     if (Object.prototype.hasOwnProperty.call(result, key)) {
-      // Check if the value is not an empty string before adding to parts
-      if (result[key] !== "") {
+      // Only add the parameter if it's not undefined, null, or an empty string
+      if (
```
HavierD, line 223: it seems like here only want a truthy `result[key]`. So the code can just be `if (result[key])` rather than 
```javascript
if (
        result[key] !== undefined &&
        result[key] !== null &&
        result[key] !== ""
      ) 

```

src/components/details/DateRangeSlider.tsx
```diff
@@ -14,6 +15,33 @@ import { DatePicker } from "@mui/x-date-pickers/DatePicker";
 import { AdapterDayjs } from "@mui/x-date-pickers/AdapterDayjs";
 import { dateDefault } from "../common/constants";
 
+const datePikerSlotProps = {
```
HavierD, line 18: duplicated code from here with `DateRangeFilter`. Please double-check the two files.

HavierD (changes requested):
A significant feature added good job! In addition, if a component is not in use and no one modify it for several months, i believe they can just be deleted directly. Others are as follow comments.

src/components/common/dropdown/RoundSelect.tsx
```diff
@@ -1,3 +1,4 @@
+// Unused component
```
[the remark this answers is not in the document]
NekoLyn: cool I will remove thoes files

src/components/common/filters/AdvanceFilters.tsx
```diff
@@ -1,48 +1,114 @@
...
+  border,
+  borderRadius,
+  color,
+  margin,
+  padding,
+  zIndex,
+} from "../../../styles/constants";
 
-export interface NonRemovableFiltersProps {
+interface FiltersProps {
```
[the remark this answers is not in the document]
NekoLyn: Thanks for reminding me! I refactored the code and changed component name, but forgot to keep it consistent with interface and file name.

src/components/common/filters/DateRangeFilter.tsx
NekoLyn: Sorry for making you confused. Since this component is developed based on DateRangeSlider(which is used in detail page, only slider), I forgot to change the component name to DateRangeFilter(which is used in filter, including a slider and a time-range-chart).

src/components/common/store/componentParamReducer.tsx
NekoLyn: That's a good point. These two methods are similar but not exactly the same. In case that we have keys like `isImosDataOnly` equls to false but we still need it appears in url for query(for now we have`isImosOnlyDataset=false`in query param), so I adopted an explicit and conservative way for checking parameters. I will keep this until we find we can abandon all the falsy values and align with ogcapi.

HavierD: 👍

src/components/common/filters/ComplexListFilter.tsx
```diff
@@ -1,3 +1,4 @@
+// Unused component
```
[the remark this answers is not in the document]
HavierD: don't forget removing this ~

HavierD (approved):
Just a minor thing you may forget. Other parts are all good for me, so i approved this PR. Could you solve the comment and then this pr can merge.

src/components/common/filters/ComplexListFilter.tsx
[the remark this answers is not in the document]
NekoLyn: woooops! done

PR #120, opened 2024-08-15
Feature/5659 change metadata info display

HavierD:
metadata contact is plain text now (not a collapsible area).
metadata dates' title was bolded and capitalized


NekoLyn (approved):
LGTM

PR #123, opened 2024-08-16
Feature/5780 empty data implementation

HavierD:
Now, if a data of a record is empty, it will show like this: 
image

this record: 0661cf33-4f86-44d7-885b-4489ca84dffe    is a good one to see the empty area


utas-raymondng (approved):
LGTM

PR #124, opened 2024-08-19
constraints of citation are open by default

HavierD:
now all the constraints on Citation panel are open by default, and still collapsible.


NekoLyn (approved):
LGTM

PR #127, opened 2024-08-22
Testing/78 test helper

utas-raymondng:
Add a test helper for playwright to access map or other object from outside of react


HavierD (approved):
A mark for automated UI testing, looks good

PR #131, opened 2024-08-26
Feature/5841 adapt search params to ogcapi

NekoLyn:


src/components/common/cqlFilters.tsx
```diff
@@ -60,9 +60,7 @@ const funcCategories: CategoriesIn = (categories: Array<Category>) => {
   const categoryLabels: string[] = [];
```
vietnguyengit, line 60: Maybe not calling the frontend variables and constants to be similar with backend too, which should be no longer "category", call them "parameterVocabLabels" for example

vietnguyengit (changes requested):

src/components/common/cqlFilters.tsx
vietnguyengit: Not just here, anywhere in the frontend that calling stuff as "category" , should all be now "parameter vocab" , make sure the plural form is correct too.

vietnguyengit: Such as "CategoriesIn" should be now "Paramet erVocabsIn" etc.

vietnguyengit: Please make sure you refer to them as "parameter vocab", not just "vocab" as there are many types of vocab, eg. "platform vocab"

utas-raymondng: I may be easier to raise a separate ticket for this as I have some code that need to merge as well, we need all ppl hold a while for this rename.

@NekoLyn can you raise another ticket?

NekoLyn: I raised a ticket about rename [#5843](https://github.com/aodn/backlog/issues/5843)

vietnguyengit: If so all good, as long as stuff will be synched.

vietnguyengit (approved):

PR #133, opened 2024-08-27
:bug: fix browser compatibility and footer

NekoLyn:
1. `fit-content` in ImageList may cause bugs in firefox browser, so need to give a correct width in ImageList
2.  add MUI `CssBaseline` which contains the  basic css base like border-box, which should be added in the project as early as possible for easier UI development


utas-raymondng (approved):
LGTM

PR #134, opened 2024-08-27
Bugfix/5763 search page pagination UI

NekoLyn:
1 Improve result list UI
2 add custom scroll bar
3 improve spider diagram, make it appears in earlier zoom level
4 in map, there is still room for improvement if needed: increase the max cluster zoom level, and check the every time zoom in if the current zoom reach the expansion zoom level of the clicked spiderified cluster to decide if need to unspiderify or not

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,18 @@
 // React import is not required in React 17 or later
 import { RouterProvider } from "react-router-dom";
-
 import Fallback from "./pages/Fallback";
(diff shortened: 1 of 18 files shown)


utas-raymondng (approved):
LGTM

utas-raymondng:
For some reason, the changes is partly there. Missing checking? The 1 - 11 is off, and the scroll bar is off too

image

PR #145, opened 2024-09-04
Feature/5754 refactor drop down select

NekoLyn:
1. refactor dropdown select
2. updated 'view' and 'sort' buttons in search result page to RC3 design
3. TODO: lift the states from Select to its parent component so select option can be memorized. But there is a case when some particular options (such as 'full map view') should not be memorized, just return a previous state. this can be improved in a separate ticket


utas-raymondng (changes requested):
Generally ok if you can fix this as well will be great

image

NekoLyn:
> Generally ok if you can fix this as well will be great
> 
> image

I am working on this ticket [#5746](https://github.com/aodn/backlog/issues/5746), will fix it and update all the other parts (except search bar since we don't have conclusion about the position of search bar yet) to RC3 design in search result page

utas-raymondng (approved):
LGTM

PR #146, opened 2024-09-06
Feature/5872 add loading icon

HavierD:
The loading modal is only used in `searchpage`. May also be useful for other pages, but not implemented in this PR. I just used a basic progress circle. They design will change in the future and at that time we only need to replace the icon.


HavierD:
some tests were also skipped for now. will fix later

src/components/loading/LoadingEmitter.ts
```diff
@@ -0,0 +1,5 @@
+import EventEmitter from "events";
+
+const loadingEmitter = new EventEmitter();
+
+export default loadingEmitter;
```
utas-raymondng, line 5: Prefer not export this emitter directly but export function that interact with it. Therefore you can enforce event type and function type for the on and off function.

utas-raymondng (changes requested):
Click the description on the list view, it shows the spatial extents but the loading keeps without exit

utas-raymondng (approved):
LGTM

PR #149, opened 2024-09-10
Disable logic to make result better

utas-raymondng:
We need a ticket to fix this search as it is confusing, basically when user type tempeature , it infer that it needs to search based on title, desc and parameter because one of the valid parameter is "temperature"


HavierD (approved):
a temp solution for a logic issue. will fix later. looks good.

PR #152, opened 2024-09-11
Bug fix on additional code cause data add twice

utas-raymondng:
Fix issue where the total count can exceed total


NekoLyn (approved):
LGTM

PR #155, opened 2024-09-11
optimize the loading

HavierD:
now the global loader is not in use, but may be useful in the future. Can keep it for several months.


utas-raymondng (approved):
LGTM

PR #158, opened 2024-09-12
optimize stability

HavierD:


src/pages/search-page/SearchPage.tsx
```diff
@@ -76,6 +77,28 @@ const SearchPage = () => {
     [_setIsLoading]
   );
 
+  const checkLoadingState = useCallback(() => {
```
utas-raymondng, line 80: If you can pass the loadingCounter as argument, then you do not need to force the function to compile again, try to pass argument to avoid re-render

utas-raymondng (changes requested):
Minor change requested

utas-raymondng (approved):
LGTM

PR #159, opened 2024-09-12
Bugfix/5895 fix layout shifting due to scrollbar disappearance

NekoLyn:
MUI components like Select, Menu, Dialog or Backdrop will create a new dom within body instead of the parent component (so that it can ignore the background when the component pop over), resulting in layout shifting due to the disappearance of the scroll bar.  
So I changed the index.css to make the scroll bar to always be visible to get rid of the layout shifting issue, which is common in web design (case of [jbhifi](https://www.jbhifi.com.au/) : when you hit any drop down select you will see there is always a placeholder for scroll bar on the right side)


utas-raymondng (approved):
LGTM

PR #164, opened 2024-09-13
:bug: fix click head menu not working

NekoLyn:
remove ClickAwayListener since it will cause popover close which is unnecessary for a hover menu


utas-raymondng (approved):
LGTM

PR #177, opened 2024-09-30
Feature/5919 links and download button function

NekoLyn:
use search params "tab" "section" if need to navigate to a particular tab or section
image


utas-raymondng (approved):
LGTM

HavierD:
While dodding, i was wondering, if `onDownload` and `onLink` functions are only used in `ResultCard`, why declare them in `SearchPage` and pass heaps of levels down? Also please check other deep-paasing functions. I was wondering, why they are declared so high, but only be used in very low levels. If they can be declared in lower level, several components' props can be much simple.

PR #180, opened 2024-10-01
Feature/5856 improve spider diagram

NekoLyn:


utas-raymondng (approved):
Look ok, but should move the code out of the component and pass argument into it when possible to avoid even the need to use call back

  // Util function to check if a cluster can spiderify or not
  const shouldCreateSpiderDiagram = useCallback(
    (features: any[]): boolean => {
      const zoom = map?.getZoom() || 0;
      const clusterCount = features[0].properties.point_count;
      return (
        (!clusterCount && features.length > 1) || zoom >= spiderifyFromZoomLevel
      );
    },
    [map, spiderifyFromZoomLevel]
  );


for example we can pass map and spidifyFromZoomLevel to function and therefore no need to keep the function inside the component.

PR #184, opened 2024-10-03
Feature/5951 decode html entities

HavierD:


NekoLyn (approved):
LGTM!
Just a small concern about the use of decode method. Since not sure if the source of abstract is trusted or let's say if it is user-generated or not, it may cause XSS attacks. You can consider the use of React's dangerouslySetInnerHTML or DOMPurify for rendering HTML content in the future.

Another concern is about the truncate of the content. For better responsive UI you can consider the prop WebkitLineClamp: "10", for example
 ```  <Typography
            sx={{
              padding: 0,
              overflow: "hidden",
              textOverflow: "ellipsis",
              display: "-webkit-box",
              WebkitLineClamp: "10",
              WebkitBoxOrient: "vertical",
               wordBreak: "break-word",
            }}
          >
            {content}
          </Typography>```
which will ensure always 10 lines are displayed and the ellipsis appeared at the end of the paragraph. 
           
This is just advice which may cause some difficulties regarding the implementation of your other functions. So I leave here in case you may need some change for this part in the future.

HavierD:
> LGTM! Just a small concern about the use of decode method. Since not sure if the source of abstract is trusted or let's say if it is user-generated or not, it may cause XSS attacks. You can consider the use of React's dangerouslySetInnerHTML or DOMPurify for rendering HTML content in the future.
> 
> Another concern is about the truncate of the content. For better responsive UI you can consider the prop WebkitLineClamp: "10", for example
> 
> ```
> <Typography
>            sx={{
>              padding: 0,
>              overflow: "hidden",
>              textOverflow: "ellipsis",
>              display: "-webkit-box",
>              WebkitLineClamp: "10",
>              WebkitBoxOrient: "vertical",
>               wordBreak: "break-word",
>            }}
>          >
>            {content}
>          </Typography>
> ```
> 
> which will ensure always 10 lines are displayed and the ellipsis appeared at the end of the paragraph.
> 
> This is just advice which may cause some difficulties regarding the implementation of your other functions. So I leave here in case you may need some change for this part in the future.


Sorry I didn't fully understand. `dangerouslySetInnerHTML` is the dangerous one, that could trigger the XSS attack. Using my decoding function properly(not with `dangerouslySetInnerHTML, or other similar operating-real-dom functions`) is safe because React uses virtual DOM, and all the texts are all be treated as pure string. e.g., if a text is `<script>dangerous code</script>`, react will only render the string : "<script>dangerous code</script>", rather than treat it as a script.

PR #187, opened 2024-10-04
some simple refactoring about search page and search functions

HavierD:
This PR shouldn't change any functions of the portal. It only contains code refactoring includes: 

- Resolve props drilling. Some functions were declared in high-level components but only used in low-level components by multiple props passing. Now some components' props are simpler and functions are only declared at where they are necessary.

### DIFF
--- a/playwright/tests/search/test_map.py
+++ b/playwright/tests/search/test_map.py
@@ -159,7 +159,7 @@ def test_map_spider(
     search_page.wait_for_search_to_complete()
 
     search_page.map.center_map(head_lng, head_lat)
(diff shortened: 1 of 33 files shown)


utas-raymondng (changes requested):
Two comments

1. Move the beforeAll().. afterAll() in all test back inside the describe() those lifecycle event need to be inside.
2. No need to put so many decimal place for the initial bbox and the polygon
3. Place the default value of the bbox and polygon in MapConfig constant.
4. I think we can remove center point param and use the bbox instead

utas-raymondng (approved):
LGTM

PR #199, opened 2024-10-22
Feature/5760 update search bar to latest UI design

NekoLyn:
1. update searchbar UI to latest design

2. remove the parameter chips of filters, only displaying a badge with number indicate number of selected filter options

3. both the 'filter' button and the 'search' button belong to the searchbar button group. They are all expandable (showing icon and button name at the same time instead of only icon) and only one button is expanded at a time in the group.

4. searchbar in landing page:
improve the key-enter-search function: do search right after press enter

5. searchbar in search result page:
only appears in header, and looks shorter than normal
will expand when there exists text input or text input field is focused
the expansion of searchbar will make the menu in header collapse into an icon (click to open)
if no text input and text input field is not focused, searchbar will shrink to the shorter one

image
image
image
image
image
image


utas-raymondng:
Reviewing it now

src/components/common/store/store.tsx
```diff
@@ -31,6 +33,8 @@ const searchQueryResult = (state: RootState) =>
 
 const getComponentState = (state: RootState) => state.paramReducer;
 
-export { searchQueryResult, getComponentState };
+const getSearchbarExpanded = (state: RootState) => state.searchbar.isExpanded;
```
utas-raymondng, line 36: Is it necessary to use store to remember state of the component? Can we determine it by text inside and then use similar approach as the map expand to signal the menu to collapse?

You can find code in map.tsx, that is you try to observer a component resize and act accordingly


      // Create a resize observer to the canvas so we know if its size have changed
      // and we need to redraw the map
      const resizeObserver = new ResizeObserver((entries) =>
        // https://stackoverflow.com/questions/70533564/mapbox-gl-flickers-when-resizing-the-container-div
        setTimeout(() => map.resize(), 0.1)
      );
      resizeObserver.observe(map.getContainer());

utas-raymondng (changes requested):
One comment on using store

And is there a ticket to followup the theme and constant color code, if yes please put a link here, else please raise one thanks

src/components/common/store/store.tsx
NekoLyn: Seems I made things complicated...I removed the change in redux and mange the search bar width in Header. Thanks!

NekoLyn:
> One comment on using store
> 
> And is there a ticket to followup the theme and constant color code, if yes please put a link here, else please raise one thanks

I will raise a ticket to refactor all the style constants in one file for next iteration. Thanks Raymond!

utas-raymondng (approved):
LGTM thanks

PR #200, opened 2024-10-23
display real data on detail page

HavierD:


src/components/map/mapbox/layers/DetailHeatMapLayer.tsx
NekoLyn, line 1: do you have a button for switching map layers in detail page map?

src/components/map/mapbox/layers/HeatmapLayer.tsx
```diff
@@ -10,21 +10,21 @@ import MapContext from "../MapContext";
...
 import SpiderDiagram from "../component/SpiderDiagram";
 import { TestHelper } from "../../../common/test/helper";
-import { FeatureCollection, Point } from "geojson";
+import { Feature, FeatureCollection, Point } from "geojson";
 import { MapDefaultConfig } from "../constants";
+import { generateFeatureCollectiionFrom } from "../../../../utils/GeoJsonUtils";
 import { mergeWithDefaults } from "../../../../utils/ObjectUtils";
 
-interface HeatmapLayer {
+interface IHeatmapLayer {
```
NekoLyn, line 27: since the HeatmapLayer contains a heatmap layer and a cluster layer in it so the config needs contain these two parts. It's an old issue can be refactor with more proper names in the future.

src/components/map/mapbox/layers/HeatmapLayer.tsx
```diff
@@ -142,7 +142,10 @@ const HeatmapLayer: FC<HeatmapLayerProps> = ({
     // to update the state and then this effect can be call again when map loaded.
     const createLayers = () => {
       const dataSource = findSuitableVisiblePoint(
-        createCenterOfMassDataSource(undefined),
+        {
+          type: "FeatureCollection",
+          features: new Array<Feature<Point>>(),
+        },
```
NekoLyn, line 148: `features`

src/components/map/mapbox/layers/ClusterLayer.tsx
```diff
@@ -98,15 +98,15 @@ export const getUnclusterPointId = (layerId: string) =>
   `${layerId}-unclustered-point`;
 
 const ClusterLayer: FC<ClusterLayerProps> = ({
-  collections,
+  features = generateFeatureCollectiionFrom(undefined),
```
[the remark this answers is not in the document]
HavierD: thank you.

src/components/map/mapbox/layers/HeatmapLayer.tsx
HavierD: i prefer using 
```
       {
          type: "FeatureCollection",
          features: new Array<Feature<Point>>(),
        },
```
rather than `features` because it is the same as the return values of previous code (always passing in `undefined`)
```
createCenterOfMassDataSource(undefined),
```

src/components/map/mapbox/layers/HeatmapLayer.tsx
HavierD: Better naming is not included in this ticket, so i just add an `I` in front as a temp solution, to make sure our project doesn't have two objects with the same name.

src/components/map/mapbox/layers/DetailHeatMapLayer.tsx
HavierD: currently no. Will do it in next iteration

src/components/map/mapbox/layers/ClusterLayer.tsx
```diff
@@ -143,15 +143,18 @@ const ClusterLayer: FC<ClusterLayerProps> = ({
       map?.addSource(clusterSourceId, {
         type: "geojson",
         data: findSuitableVisiblePoint(
-          createCenterOfMassDataSource(undefined),
+          {
+            type: "FeatureCollection",
+            features: new Array<Feature<Point>>(),
+          },
```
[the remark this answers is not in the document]
HavierD: i prefer not using `features` to be consistent with previous code

src/components/map/mapbox/layers/HeatmapLayer.tsx
NekoLyn: yes I understand you using a empty FeatureCollection here. but in destructuring props you give a default empty FeatureCollection<Point> here `features = generateFeatureCollectionFrom(undefined)` if feature is undefined, and 
```
export interface LayersProps {
  // Tile layer should added to map
  // collections?: Array<OGCCollection>;
  features?: FeatureCollection<Point>;
  // Event fired when user click on the point layer
  onDatasetSelected?: (uuids: Array<string>) => void;
  // dataset that user selected from result list or map
  selectedUuids?: string[];
  showFullMap?: boolean;
  tabNavigation?: (uuid: string, tab: string, section?: string) => void;
}
```
also ensure the passed in feature (if has) is FeatureCollection<Point>, so I think no need to write an empty FeatureCollection here can use `features` directly

in previous code, it didn't give default empty initial value in props destructure, therefore using `createCenterOfMassDataSource(undefined)` here to make sure there will be an empty FeatureCollection<Point>. so you can also use ` generateFeatureCollectionFrom(undefined)` for getting the empty FeatureCollection<Point> since you have already used it in previous code.

NekoLyn (approved):
LGTM

PR #203, opened 2024-10-28
:sparkles: improve clear text 'x' button

NekoLyn:
before change: clear text 'X' button only appears when hovering on search bar input text field

after change:  clear text 'X' button is showing when hovering on text field or there exists input text.


HavierD (approved):
LGTM

PR #204, opened 2024-10-28
:bug: fix map hover tip

NekoLyn:
Mapbox creates popup outside the react dom tree, therefore the popup and its child components are outside the react router context. so I pass the tabNavigation method down to the popup component.
Also refactor ResultCardButtonGroup to make button functions can be passed in so that the component is more reusable in different contexts.


HavierD (approved):
LGTM

PR #215, opened 2024-11-13
Feature/6047 improve search page map function

NekoLyn:


NekoLyn:
TODO:
1. optimize the expand/collapse transition of the pin list accordion group, to make it running smoothly 
2. implement the pin list card function like the button functions
3. any styles need to improve like colors and icons
4. improve the mini map: sometimes it refresh by itself and lose the original bound and zoom; need to disable the animation

NekoLyn:
done in https://github.com/aodn/aodn-portal-v2/pull/216

PR #223, opened 2024-11-20
Feature/6041 detail map layer button

HavierD:
use aggregated data;
frontend is now able to draw polygons on detail page map.


utas-raymondng (approved):
LGTM

PR #224, opened 2024-11-20
:sparkles: implement full list view and fetch more

NekoLyn:
a simple full list view, need to improve styles and implement button functions, can raise a follow up ticket according to feedback

TODO:
1. need pin list? if need, should put in higher level so pin list can be remembered after switching view mode or pages
3. bug: default fetch page size is 11 when changing sort options in full list view

image

image


src/pages/search-page/SearchPage.tsx
```diff
@@ -427,35 +427,35 @@ const SearchPage = () => {
         }}
         gap={2}
       >
-        {/* Ignore the FULL_LIST view for now, wait for future design */}
-        {selectedLayout !== SearchResultLayoutEnum.FULL_MAP && (
-          <Box>
-            <ResultSection
-              onClickCard={onClickCard}
+        <Box>
```
utas-raymondng, line 430: I prefer follow the same patter then the ResultSection have an attribute like below showFullMap, where you can showFullList

src/components/result/FullListAndDetails.tsx
```diff
@@ -0,0 +1,80 @@
...
+      const paramPaged = createSearchParamFrom(componentParam, {
+        pagesize: FULL_LIST_PAGE_SIZE,
+      });
+      dispatch(fetchResultWithStore(paramPaged));
+    }
+    //only for fetching initial data
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const fetchMore = useCallback(async () => {
```
utas-raymondng, line 48: I think the fetch is done on the top level so you do not need to fetch it here?

utas-raymondng (changes requested):
Lets have a quick chat on some of these changes

src/components/result/FullListAndDetails.tsx
NekoLyn: thanks! I created a useFetchData hook for reusing in different components

src/pages/search-page/SearchPage.tsx
NekoLyn: I modified to pass showFullList and showFullMap

utas-raymondng (approved):
LGTM

PR #225, opened 2024-11-21
Feature/6041 detail map layer button

HavierD:
show dots on detail map;
be able to draw polygon;
show data summary in a popup when clicking a dot


NekoLyn (approved):
LGTM

PR #226, opened 2024-11-22
Bugs/6008 fix missing centroid

utas-raymondng:
Fix issue on calculate if centroid is visible across meridian that is -203 for lng etc

You can test it by finding a spatial extents that is close to 180 <> -180 line and then move the map to right but still able to see the spatial extents. The dot should still fail inside the spatial extents


NekoLyn (approved):
LGTM

PR #229, opened 2024-11-27
Fix test with warning act and using BBOX

utas-raymondng:
Migrate to use BBOX and fix test error on requiring act()

basically you should not use await with user click in test


HavierD (approved):
LGTM

PR #230, opened 2024-11-27
Bugs/6119 avoid duplicate spatial call

utas-raymondng:
Remove one extra unnecessary load by not setup a event handler
Move the json to public to avoid it bundle with the initial load


NekoLyn (approved):
LGTM

PR #234, opened 2024-12-04
Update to include record without spatial extents

utas-raymondng:
This allow record without spatial extents show on the list for example

"IMOS - ANMN National Reference Stations - Darwin and Yongala - Near real-time meteorology and oceanographical data"

Change related to this ticket result -> https://github.com/aodn/backlog/issues/5953


NekoLyn (approved):
LGTM

PR #235, opened 2024-12-06
Feature/6104 rectangle time control

HavierD:
As the previous meeting, we implement the core functions first and then do the fancy thing. So the most of the style values are hard coded now as temporary solution.


src/components/box/BBoxItem.tsx
```diff
@@ -0,0 +1,26 @@
...
+  label: string;
+  value: number;
+}
+
+const BBoxItem: React.FC<BBoxItemProps> = ({ label, value }) => {
+  const roundedValue = _.round(value, 2);
+  return (
+    <Typography
+      sx={{
+        color: "#7194AB",
```
utas-raymondng, line 15: Use color constant or styled component?

src/components/box/DateRangeConditionBox.tsx
```diff
@@ -0,0 +1,40 @@
...
+  DownloadConditionType,
+} from "../../pages/detail-page/context/DownloadDefinitions";
+import React, { useMemo } from "react";
+import { Typography } from "@mui/material";
+
+interface DateRangeConditionBoxProps {
+  dateRangeCondition: DateRangeCondition;
+}
+
+const DateRangeConditionBox: React.FC<DateRangeConditionBoxProps> = ({
```
utas-raymondng, line 13: Try this http://localhost:5173/details?uuid=06b09398-d3d0-47dc-a54a-a745319fbece&tab=abstract

I cannot move the dot and select time range, end date is 10-1800

src/components/box/DownloadConditionBox.tsx
```diff
@@ -0,0 +1,101 @@
...
+      </Grid>
+      <Grid container item md={8}>
+        <Grid
+          sx={{ paddingX: theme.mp.lg, paddingY: theme.mp.sm }}
+          item
+          md={12}
+        >
+          <Typography
+            sx={{
+              color: "#7194AB",
```
utas-raymondng, line 75: Again styled component or constant color?

src/components/details/TabsPanelContainer.tsx
```diff
@@ -1,10 +1,9 @@
-import Tab from "@mui/material/Tab";
 import Box from "@mui/material/Box";
 import { padding } from "../../styles/constants";
 import StyledTabs from "../common/tab/StyledTabs";
 import StyledTab from "../common/tab/StyledTab";
 import { useLocation } from "react-router-dom";
-import { FC, SyntheticEvent, useEffect, useState } from "react";
+import React, { SyntheticEvent, useEffect, useState } from "react";
```
utas-raymondng, line 6: Not related to your change, but there is a function from Lyn call allyProps, can change it to const allyProps = (index:number) => ....   which align with other function signature?

package.json
```diff
@@ -47,6 +47,7 @@
     "http-proxy-middleware": "^2.0.6",
     "lodash": "^4.17.21",
     "mapbox-gl": "3.4.0",
+    "mapbox-gl-draw-rectangle-mode": "^1.0.4",
```
utas-raymondng, line 50: I see you use the DrawRectangle.js, do you still need to import this or you just import this and delete the DrawRectangle.js?

src/components/map/mapbox/controls/DateRangeControl/DateRangeControl.tsx
```diff
@@ -0,0 +1,95 @@
...
+    this.iconRoot.render(
+      <button>
+        <img alt="" src={timeRange}></img>
+      </button>
+    );
+    this.container.onclick = this.setIsShowingDateSelector;
+    return this.container;
+  }
+
+  onRemove(_: mapboxgl.Map) {
```
utas-raymondng, line 37: Please refer to ToggleControl.tsx and how to unmount and discard control properly, if you can refactor to make it common so that all control call the same discard will be even better

src/components/map/mapbox/controls/DateRangeControl/DateSliderControl.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <DateSlider
+        minDate={this.minDate}
+        maxDate={this.maxDate}
+        onDateRangeChange={this.onDateRangeChange}
+      />
+    );
+    return this.container;
+  }
+
+  onRemove() {
```
utas-raymondng, line 64: Same as above on how to discard it correctly

src/components/map/mapbox/controls/DrawControl.tsx
```diff
@@ -0,0 +1,103 @@
...
+  map,
+  setDownloadConditions,
+  draw,
+}) => {
+  const [open, setOpen] = useState<boolean>(false);
+
+  const anchorRef = useRef(null);
+  const popperRef = useRef<HTMLDivElement>(null);
+
+  const updateArea = useCallback(() => {
```
utas-raymondng, line 35: Can you put the updateArea inside the on off event block so that it didn't need to use callback?

src/components/map/mapbox/controls/DrawRectangleControl.tsx
```diff
@@ -0,0 +1,39 @@
...
+    this.container.style.width = "29px";
+    this.container.style.height = "29px";
+    this.container.style.display = "flex";
+    this.container.style.alignItems = "center";
+    this.container.style.justifyContent = "center";
+
+    return this.container;
+  }
+  onRemove() {
+    this.container?.parentNode?.removeChild(this.container);
```
utas-raymondng, line 34: Same as above on remove control

src/pages/detail-page/context/detail-page-provider.tsx
```diff
@@ -27,6 +48,26 @@ export const DetailPageProvider: FC<DetailPageProviderProps> = ({
   >(undefined);
   const [isCollectionNotFound, setIsCollectionNotFound] =
     useState<boolean>(false);
+  const [downloadConditions, _setDownloadConditions] = useState<
```
utas-raymondng, line 51: _set instead of setDownloadConditions is because?

utas-raymondng (dismissed):
If I draw two rectangle, how to tell which one I am going to delete?

And other comments

src/components/box/BBoxItem.tsx
HavierD: Because there is already a new design update : https://github.com/aodn/backlog/issues/5991#issuecomment-2516246028 and i was told the design will also be discussed and might change further, in addition the #7194AB is a new color, so i hard coded it and will change according to new design in the future.

HavierD:
> If I draw two rectangle, how to tell which one I am going to delete?
> 
> And other comments

all condition class implements IDownloadCondition interface so every condition must provide an id when initializing it. for the rectangle condition, the mapbox provided a uuid for every bbox. So i used these id to identify different bbox.

src/pages/detail-page/context/detail-page-provider.tsx
HavierD: because i wanted to customize the `setstate` function. So i used `_setxxx` as the `setState` function, and customized it as: `setxxx` which invokes the `_setxxx` to rerender related components.

package.json
HavierD: I think we can keep both. The reason why i did this is, importing it in `package.json` is to make it available to dependabot. Leveraging dependabot is easier for us to know whether this `mapbox-gl-draw-rectangle-mode` is safe. If it was alerted by dependabot, we can then directly change the DrawRectangle.js to fix it manually.

src/components/map/mapbox/controls/DateRangeControl/DateRangeControl.tsx
HavierD: 👌  thank you. On it

NekoLyn (approved):
LGTM

PR #236, opened 2024-12-08
Feature/6095 bookmark list

NekoLyn:


src/components/map/mapbox/component/MapPopup.tsx
```diff
@@ -19,8 +19,9 @@ import AppTheme from "../../../../utils/AppTheme";
 import { OGCCollection } from "../../../common/store/OGCCollectionDefinitions";
 import { useAppDispatch } from "../../../common/store/hooks";
 import ComplexMapHoverTip from "../../../common/hover-tip/ComplexMapHoverTip";
+import { BookmarkButtonBasicType } from "../../../bookmark/BookmarkButton";
 
-interface MapPopupProps {
+interface MapPopupProps extends Partial<BookmarkButtonBasicType> {
```
utas-raymondng, line 24: Do not make sense to extends Partial<BookmarkButtonBasicType>

src/pages/search-page/SearchPage.tsx
```diff
@@ -315,86 +315,113 @@ const SearchPage = () => {
...
         setSelectedUuids([item.id]);
-        setSelectedUuid(item.id);
       }
     },
-    [fillGeometryIfMissing]
+    [dispatch]
+  );
+
+  // For checking if the uuid of an item is in bookmarkItems list
+  const checkIsBookmarked = useCallback(
```
utas-raymondng, line 346: You do not need to define checkIsBookmarked and onClickBookmark here and pass it around, you can just define the function in reducer and remove all those passing props. Then you add a emitter inside the bookmarkListReducer and give a on off event subscribe, any change of the bookmark is done via function exposed by this reducer dispatch(addBookmark(uuid)) or dispatch(removeBookmark(uuid)), inside the add function not only you update the localstorge but then fire event to each subsbscriber via the on function so that the button can be toggle automatcially.

utas-raymondng (changes requested):
Please make changes to store first and remove all the passing props, once the code is clean a bit, we can give it another round of PR.

src/pages/search-page/SearchPage.tsx
utas-raymondng: Inside the reducer of the bookmark, 

1. const emitter = new Emitter(); // Inside reducer
2. const on = (string (event), handle) => emitter.on(string, handler);
3. const off = (string (event), handle) => emitter.off('xxx', handle);
4. const addBookmark = (uuid) => { save to localstore && emitter.emit('ADD_BOOKMARK', handle);
5. const removeBookmark = (uuid) => { delete localstore item && emitter.emit('REMOVE_BOOKMARK', handle);
6. const checkIsBookmark = (uuid) => { return true if bookmark set in localstorr or redex)


Once you done that, remove all the props of checkIsBookmarked and onHandleBookmarks

For component call checkIsBookemarked -> dispatch(checkIsBookmarked(uuid))

PR #243, opened 2024-12-17
Feature/6059 update search bar

NekoLyn:
1. skipped some failed playwright tests due to change of the new search bar and filter ( followup https://github.com/aodn/aodn-portal-v2/issues/244)
2. hide some filter sections and wait for ogcapi
3. temporarily combine the platform filter into text search


utas-raymondng (approved):
LGTM

PR #248, opened 2024-12-24
Feature/6120 bookmark list in full list view

NekoLyn:


utas-raymondng (approved):
LGTM

src/components/map/mapbox/controls/menu/MenuControl.tsx
```diff
@@ -112,6 +117,13 @@ const MenuControl: React.FC<MenuControlProps> = ({
     });
   }, [map, menu]);
 
+  // Props update effect
```
utas-raymondng, line 120: This useEffect will reset the behavior and cause some menu item not work, do not use the useEffect here, the PR overlook this and will fix with another PR because code merged

PR #250, opened 2025-01-01
Fix bug on BookmarkListMenu.tsx after remove useEffect

utas-raymondng:
Cannot put useEffect inside the MenuControl as it will impact the other control on detail page


NekoLyn (approved):
LGTM

PR #249, opened 2025-01-01
Playwright: Fix skipped tests

AR-RIFAT:
Updated the data test ID for the search button in Playwright tests.

PR #251, opened 2025-01-03
Features/6116 spatial map focus detail map

utas-raymondng:


utas-raymondng:
Didn't aware that you put a fix in 

https://github.com/aodn/aodn-portal-v2/pull/249

NekoLyn (approved):
LGTM

PR #260, opened 2025-01-12
Feature/6172 responsive layout landing page

NekoLyn:


utas-raymondng:
Found two issues

1.  Once you click the dropdown on top, it will be there forever
2. The button of the search bar drop out, not sure what is the right design
image

utas-raymondng:
> Found two issues
> 
>     1. Once you click the dropdown on top, it will be there forever
> 
>     2. The button of the search bar drop out, not sure what is the right design
>        image

Ignore above
1. This due to the need to select the touch button on the web browser to simulate the finger operation
2. Need @quye04 update design on where the button goes

utas-raymondng (approved):
LGTM

PR #266, opened 2025-01-14
Playwright: test map zoom out and drag crash

AR-RIFAT:


utas-raymondng:
Question, how to tell if the map crashes or not in the test after the last operation?

utas-raymondng (approved):
LGTM

PR #267, opened 2025-01-17
Feature/5266 search result page responsiveness

NekoLyn:


utas-raymondng (approved):
LGTM, please use another PR to address these comments

1. may be we should default to full list view by default on mobile. (need comment on end of iteration review)
2. For full map view the map cut on iphoneSE, I guess need to use vh etc instead of hardcode?
3. Also zoom level need to change so that user still see the whole Auz on small screen

PR #276, opened 2025-01-23
Limited number of record return on map

utas-raymondng:
It takes too long for search without any search criteria, it is better to limited the number of record return to speed up the system.


HavierD (approved):
LGTM

PR #280, opened 2025-01-24
Bugs/6229 return button on detail page

utas-raymondng:
if you go to detail page directly, the return button will back to landing page


HavierD (approved):
LGTM

PR #286, opened 2025-02-04
Playwright: test detail page map download filter

AR-RIFAT:
1. Modified `TestHelper` to handle multiple maps in the test framework
2. Added test to verify that drawing a rectangular shape on the detail map creates a download filter item.
3. Added test to verify that selecting a date range via the slider creates a download filter item.


utas-raymondng (approved):
LGTM

PR #294, opened 2025-02-12
Feature/minor updatingq

HavierD:
1. capital letters in email won't bother the system anymore.
2. if a process request is taking too long, it won't spin infinitely. It will show timeout message now.


utas-raymondng (approved):
LGTM

PR #300, opened 2025-02-14
Features/6225 location filter

utas-raymondng:
Add location filter, the query generated will include BBOX and INTERSECT, however the OGC api seems report something wrong where if you select a location which is in Australia and then you drag the map to some other place, you should expect no record but you still get records

A separate PR on OGC api is expected


NekoLyn (approved):
LGTM

PR #319, opened 2025-02-27
Bugfix/6365 fix map bottom popup card

NekoLyn:
change in CardPopup:
for click event, if it include a` targetLayerId`, we will use this layerId for later query

in spider diagram layer:
the click function will also fire a map click with custom event (has `targetLayerId` included)

other changes:
update ResultCardButtonGroup style and `onDownload` function


utas-raymondng (approved):
LGTM

PR #324, opened 2025-03-04
:bug: fix fit to bound

NekoLyn:
create fitToBound map util function for fitting the map view to the specified bounding box with intelligent zoom calculation


HavierD (approved):
LGTM

PR #334, opened 2025-03-18
:bug: fix uncluster layer remove

NekoLyn:


HavierD (approved):
found an issue while querying data from ogc-api edge, but seems like something wrong with the cloudfront config, and nothing to do with this ticket. LGTM

PR #335, opened 2025-03-19
Playwright: configured playwright tracing

AR-RIFAT:


playwright/core/factories/device_configs.py
```diff
@@ -0,0 +1,18 @@
+from playwright.sync_api import Playwright
+
+from core.dataclasses.device_config import DeviceConfig
+
+
+def get_mobile_config(playwright: Playwright) -> DeviceConfig:
+    iphone_se = playwright.devices['iPhone SE']
+    return DeviceConfig(
+        is_mobile=True,
+        viewport={'width': 375, 'height': 667},
```
HavierD, line 10: could you extract these size number constants into a constants file? that may be helpful for the future if there are more sizes or other constants coming.

playwright/core/factories/device_configs.py
```diff
@@ -0,0 +1,18 @@
...
+        device_config=iphone_se,
+    )
+
+
+def get_desktop_config() -> DeviceConfig:
+    return DeviceConfig(
+        is_mobile=False, viewport={'width': 1920, 'height': 1080}
```
HavierD, line 17: also number constants here

HavierD (commented):
a small comment

playwright/core/factories/device_configs.py
AR-RIFAT: Thanks for your feedback. I have moved the device-related constants into `core/constants/devices.py` where they are now organized by device type. In the future, if needed, it will be easier to add new devices using this setup. Let me know if you'd like any other adjustments.

HavierD (approved):
LTGM

PR #343, opened 2025-03-31
Bugfix/6314 save time period selection

NekoLyn:
- a helper function is created to check if the given time range falls into any predefined year range (from the radio group)
- `const TOLERANCE_DAYS = 3` - tolerance in days (can adjust if needed) is applied for matching predefined date ranges


utas-raymondng (approved):
LGTM

PR #346, opened 2025-04-02
Bug/6409 time range fix

HavierD:
- symbol popup shows correct information;
- date range selector on map is working properly now.


src/components/map/mapbox/controls/menu/MenuControl.tsx
```diff
@@ -23,7 +23,7 @@ interface MenuControlProps {
 class MapControl implements IControl {
   private container: HTMLDivElement | null = null;
   private root: Root | null = null;
-  private readonly component: MapControlType;
+  private component: MapControlType;
```
utas-raymondng, line 26: This must be readonly, if you check previous commit, Lyn make similar changes and break other area, please rollback changes of this file

utas-raymondng (changes requested):
Please rollback changes to this file before continue

PR #347, opened 2025-04-02
Fix bug on copy link button and remove no use button

utas-raymondng:
If you go to detail page and find a link tab, only one Copy Link button will have copy status, if you click another button it reset the copied status, however there is another issue with the show of the button which is outside the scope of this ticket. https://github.com/aodn/backlog/issues/6610


NekoLyn (approved):
LGTM

PR #348, opened 2025-04-03
Feature/6556 improve topic cards scroll

NekoLyn:
image
image
image
image
image


utas-raymondng (approved):
LGTM  but can we use the debounce in other places?

NekoLyn:
> LGTM but can we use the debounce in other places?

yes can use the useDebounce hook anywhere

PR #353, opened 2025-04-04
Bugfix/6609 fix multiple copy link button

NekoLyn:
HOVER
image
CLICK
image
SYNC
image


NekoLyn:
TODO: need to consider the touch screen, which doesn't have hover event. So better to always have a copy icon showing next to the text that can be copied. 
example:
image

utas-raymondng:
Agree may be we should not hide the copy button in the first place, what you think @quye04

PR #355, opened 2025-04-08
Fix popup location

utas-raymondng:
The problem is the content is not completely load before popup shown, therefore the location of the popup is not correct


NekoLyn (approved):
LGTM

PR #357, opened 2025-04-08
Features/6439 hexbin map layout

utas-raymondng:
The speed is not great but should be ok for initial presentation


utas-raymondng:
Will try another approach using deck.gl

PR #359, opened 2025-04-10
Bugfix/6239 mini map in bookmarks

NekoLyn:
BUG:
 the bug is caused by missing `extent` in collection object when add bookmark from the result cards/ in bookmark initialization;
so added `bbox` in search param properties for query to endpoint `/api/v1/ogc/collections`

TEST:
1. adding bookmark from result card/map hovertip, and expanding it from bookmark list to see if the mini map showing extents correctly
2. adding some bookmarks, refresh the page and see if mini map in bookmark card showing extents correctly

TODO: 
found there exists records (eg. 37662a66-ec3a-4ece-9db0-31930c181725) with NO extents, need to investigate. if it is normal need to consider if we need hide mini map or replace with pic says there is no extent?

```
[
    {
        "properties": {
            "status": "completed"
        },
        "id": "37662a66-ec3a-4ece-9db0-31930c181725",
        "title": "Slocum glider measurements of the Australian North West Shelf in April 2012",
        "description": "A total of four gliders were deployed on the Australian North West Shelf from April 1-21, ....",
        "links": [ ...  ],
        "extent": {
            "spatial": {
                "crs": "http://www.opengis.net/def/crs/OGC/1.3/CRS84"
            },
            "temporal": {
                "trs": "http://www.opengis.net/def/uom/ISO-8601/0/Gregorian"
            }
        },
        "itemType": "Collection"
    }
]
```

### DIFF
--- a/src/components/bookmark/BookmarkButton.tsx
+++ b/src/components/bookmark/BookmarkButton.tsx
@@ -5,7 +5,7 @@ import { IconButton, Tooltip } from "@mui/material";
 import { color, gap } from "../../styles/constants";
 import { OGCCollection } from "../common/store/OGCCollectionDefinitions";
 import {
-  addItem,
+  checkExtentAndAdd,
   checkIsBookmarked,
   getBookmarkList,
   off,
@@ -52,7 +52,7 @@ const BookmarkButton: FC<BookmarkButtonProps> = ({
           // If bookmark a temporary item, should clear temporaryItem then add to bookmark list
           store.dispatch(setTemporaryItem(undefined));
         }
-        store.dispatch(addItem(item));
+        store.dispatch(checkExtentAndAdd(item));
       }
       setIsBookmarked(!isBookmarked);
     },
--- a/src/components/bookmark/BookmarkButton.tsx
+++ b/src/components/bookmark/BookmarkButton.tsx
@@ -4,8 +4,9 @@ import BookmarkIcon from "@mui/icons-material/Bookmark";
 import { IconButton, Tooltip } from "@mui/material";
 import { color, gap } from "../../styles/constants";
 import { OGCCollection } from "../common/store/OGCCollectionDefinitions";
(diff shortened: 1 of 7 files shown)


src/components/common/store/bookmarkListReducer.tsx
```diff
@@ -141,12 +140,41 @@ export const {
...
+      })
+      .catch((err: Error) => {
+        errorHandling(thunkAPI);
+      });
+  }
+});
+
 // Thunk actions for async operations, avoid getState() where someone is calling the reducer
-export const fetchAndInsertTemporary = createAsyncThunk<
+export const checkAndInsertTemporary = createAsyncThunk<
```
utas-raymondng, line 170: Do not like to do it this way, it is better to have the caller component to sort that out for you instead of backfill something here, same for checkAndInsertTemporary, can you redo without adding two functions?

utas-raymondng (changes requested):
Request change without create new check function

src/components/bookmark/BookmarkButton.tsx
```diff
@@ -19,6 +20,7 @@ import {
   BookmarkEvent,
   EVENT_BOOKMARK,
 } from "../map/mapbox/controls/menu/Definition";
+import { checkExtent } from "../../utils/GeoJsonUtils";
```
utas-raymondng, line 23: No good to still keep the query , is this the function that create the collection?

    const getCollectionData = useCallback(
      async (uuid: string) => {
        return dispatch(fetchResultByUuidNoStore(uuid))
          .unwrap()
          .then((collection: OGCCollection) => collection)
          .catch((error: any) => {
            console.error("Error fetching collection data:", error);
            // TODO: handle error in ErrorBoundary
          });
      },
      [dispatch]
    );

utas-raymondng (changes requested):
The initial query should have all the info need before rendering happens

utas-raymondng (approved):
LGTM

PR #360, opened 2025-04-10
Bugfix/6409 time range update

utas-raymondng:
Create a new function so that component receive props update, I have not fix the bug where slider change do not result in correct data filter, can @HavierD help ?


HavierD (approved):
LGTM

PR #363, opened 2025-04-15
Upgrade package version, avoid security issue and fix a typo in map

utas-raymondng:
There is a typo in zoomstart , where it is type as zoomestart. Move the value to constant and optimized the map with memo plus upgrade package version to avoid security issue


NekoLyn (approved):
LGTM

PR #367, opened 2025-04-23
Features/6439 hexbin layer deckgl

utas-raymondng:
This is a faster version with deckgl


HavierD (commented):
a couple of minor suggestion. Thank you

HavierD:
my another concern is, if the dataset is small, it is very hard to find them on the map. Maybe we can use more obvious color, or other ways to deal with it. For example this dataset: 78d588ed-79dd-47e2-b806-d39025194e7e 
image

We may need to discuss it later.

utas-raymondng:
> my another concern is, if the dataset is small, it is very hard to find them on the map. Maybe we can use more obvious color, or other ways to deal with it. For example this dataset: 78d588ed-79dd-47e2-b806-d39025194e7e image
> 
> We may need to discuss it later.

Yup need discussion

HavierD (approved):
LGTM

PR #372, opened 2025-05-02
feature/6585-add-test-for-view-sort-memory

NekoLyn:
Added/combined some testcases to test basic flow in search page
for now tests including:
test 1: map expand properly
test 2: load correct record when click load more button 
test 3: URL parameters to Redux state flow
test 4: Redux state to UI flow
test 5: layout button click to Redux and URL parameters flow

since there still some bugs in `SearchPage` so will add more tests in the future after bug fixed

TODO: found that many existing test may be falsely positive tests when using `waitFor` without return it. So we need to double check the current test cases and fix them if needed.


utas-raymondng (approved):
LGTM

PR #377, opened 2025-05-09
Feature/6577 add geoserver layer

NekoLyn:
Example uuid=ae86e2f5-eaaf-459e-a405-e654d85adb9c
- GeoServer Layer
image

-  Zoom in
image

- Layer switcher button
image

- No WMS data / get GeoServer tile resources failed
image

### DIFF
--- a/src/components/map/mapbox/layers/HexbinLayer.tsx
+++ b/src/components/map/mapbox/layers/HexbinLayer.tsx
@@ -118,9 +118,9 @@ const HexbinLayer: FC<LayerBasicType> = ({ featureCollection }) => {
   );
 
   useEffect(() => {
-    if (map === null || overlayRef.current) return;
+    if (!map || overlayRef.current) return;
 
-    const createHexbinLayer = (map: Map) => {
+    const createHexbinLayer = () => {
       const overlay = createLayer(featureCollection, map);
       if (overlay) {
         overlayRef.current = overlay;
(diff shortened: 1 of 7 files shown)


NekoLyn:
TODO: 
Currently getting geoserver wms tile resources is  implemented by proxy:
```
 "/geowebcache": {
          target: "https://tilecache.aodn.org.au",
          changeOrigin: true,
          secure: false,
        },
```
we may need to same proxy configuration in our deployment infrastructure to ensure CORS headers are properly set.

src/components/map/mapbox/layers/HexbinLayer.tsx
```diff
@@ -120,24 +120,28 @@ const HexbinLayer: FC<LayerBasicType> = ({ featureCollection }) => {
...
       if (overlay) {
         overlayRef.current = overlay;
         map?.addControl(overlay);
       }
-    });
-    map?.once("remove", () => {
+    };
+
+    map?.once("load", () => createHexbinLayer(map));
+    map?.on("styledata", () => createHexbinLayer(map));
```
utas-raymondng, line 132: You create two individual function here for on off, which is not what you want, you should do 

const f = () => createHexbinLayer(map!)

then on("styledata", f)

off("styledata", f)

This will fix the problem where the hexlayer still appear when you switch layer

utas-raymondng (changes requested):
Minor fix needed

src/components/map/mapbox/layers/HexbinLayer.tsx
NekoLyn: thanks! bug fixed

utas-raymondng (approved):
LGTM

PR #379, opened 2025-05-13
Feature/6620 small screen header

NekoLyn:
image
image
image


HavierD (changes requested):
could you add a simple vitest cases for the new component headerIconMenu? just several simple behaviours are enough e.g., after clicking "about us", show the menu etc...

HavierD (approved):
LGTM

PR #385, opened 2025-05-20
fix missing contacts

HavierD:


HavierD:
the skipped test case will be fixed [here](https://github.com/aodn/aodn-portal-v2/issues/387)

NekoLyn (approved):

PR #386, opened 2025-05-20
:bug: fix search page reload issue

NekoLyn:
1. use URL as the single source of truth to update redux once page loaded, which is good for page reload/ URL sharing 
2. search param memory logic: when back to landing page, all the search params are cleared, user can start a very new search from landing page; in other cases search params keep consistent across website
3. also fixed the bugs  https://github.com/aodn/backlog/issues/6689 https://github.com/aodn/backlog/issues/6691


utas-raymondng (approved):
LGTM

PR #391, opened 2025-05-27
Playwright: add tests for search related UI interactions

AR-RIFAT:


utas-raymondng:
@Nirali-IMOS maybe you can take a look as well thanks

utas-raymondng:
If you an provide some description on the test that it is doing will be good

PR #394, opened 2025-06-01
Feature/6737 change detail page url pattern

NekoLyn:
image


NekoLyn:
skipped playwright tests for detail page due to the change of URL pattern, ticket raised https://github.com/aodn/aodn-portal-v2/issues/395

utas-raymondng (approved):
LGTM, you need to raise ticket to Ashraful to update the test thanks

PR #396, opened 2025-06-02
Bugfix/6582 date range filter optimize test

utas-raymondng:
I spot an error related to the serialization of of object in Redux during test, it may not show in normal run but that signal an potential issue. So I fix it in the PR too.

Basically Redux expect action.payload to be something that can be serializable, for example string. However before the change the payload is OGCCollections with method that cannot be serializable, therefore warning/error appears in test run for data range filter.

I changed the function to return string and move the conversion to the reducer part. For the no store method, caller is expect to call the conversion after obtain the string.


NekoLyn (approved):
LGTM

PR #404, opened 2025-06-05
:bug: convert bookmark initialization to OGCCollections

NekoLyn:


utas-raymondng (changes requested):
Click the bookmark on the list, blue ribbon is ON, then refresh, the bookmark works but the blue ribbon do not set on the list view

PR #417, opened 2025-06-25
Fix changes on zoom to selected area

utas-raymondng:
Remove the need to show searching on map area because the doMapSearch is async so it will not block anyway

Also fix issue on when to apply zoom on selected area


NekoLyn (approved):
LGTM

PR #418, opened 2025-06-26
Bugfix/6431 fix bug on focus

utas-raymondng:
Fix bug on focus
Optimized the useMemo usage
Add memo to component
Fix bug where the fire event will potentially cause infinity loop


HavierD (approved):
LGTM

PR #433, opened 2025-07-11
Allow call cancel to avoid flip with unmount

utas-raymondng:
Sometime the map still flick, when you move the map around during search and it sometime recenter, this is due to the call to doMapSearch with  needNavigate. 

Since this is something existing , it is better resolve it with different ticket


HavierD (approved):
LGTM

PR #435, opened 2025-07-14
Feature/7035 figma color library integration

utas-amberbian:


src/styles/themeRC8.ts
```diff
@@ -0,0 +1,267 @@
...
+    heading2: true;
+    heading3: true;
+    heading4: true;
+    title1Medium: true;
+    title2Regular: true;
+    body1Medium: true;
+    body2Regular: true;
+    body3Small: true;
+  }
+}
```
NekoLyn, line 159: if we can separate type declaration and theme creation would be better
and use .d.ts for type declaration file which is easier for identification

recommend structure:
├── styles/
│   ├── designTokens.ts        
│   ├── theme.ts               
│   └── types/
│        └── typography.d.ts  
│        ├── palette.d.ts  
│        └── other type declaration.d.ts
│...

NekoLyn (commented):
Looks good. Just some file structure change

NekoLyn (approved):
LGTM

PR #445, opened 2025-07-18
Bugfix/7098 too many history update

utas-raymondng:
Also revert a change where lz-string on URL will not work with AWS as the char seems chopped. Fall back to string base and encode with simple replacement

branch used incorrect number should be 7093


NekoLyn (approved):
LGTM

PR #448, opened 2025-07-20
Fix typescript version error and init map search map fail

utas-raymondng:
Fix an issue where init map search may be cancel incorrectly


NekoLyn (approved):
LGTM

PR #457, opened 2025-07-24
Fix issue with search button and add test

utas-raymondng:
Problem is the search click do not trigger useEffect update due to using wrong dependency


NekoLyn (approved):
LGTM

PR #467, opened 2025-07-30
Bugfix/wcag showing result short view fonts 7057

utas-amberbian:


NekoLyn:
good work just a minor UI issue:
image

PR #480, opened 2025-08-04
:bug: fix info tip email link

NekoLyn:
- add link check for info card content (email and url)
- fix LinkCard height


utas-amberbian (approved):
LGTM

PR #493, opened 2025-08-20
Bugfix/7067 wcag bookmark header

utas-amberbian:


NekoLyn (approved):
LGTM

NekoLyn:
The result card button group component and result card component will be refactor in the future to align with the latest design
so this pr looks good to me, I merge this pr for quick fix

PR #540, opened 2025-09-16
Feature/6946 track copy citation click events

utas-amberbian:


src/analytics/customEventTracker.ts
```diff
@@ -0,0 +1,13 @@
+import { TrackingEventName } from "./constants";
+
+export function trackCustomEvent(
+  eventName: TrackingEventName,
+  eventParameters: Record<string, any> = {}
+) {
+  console.log("Custom Event:", eventName);
+  console.log("Event Parameters:", eventParameters);
+
+  if (window.gtag) {
```
utas-raymondng, line 10: Short cut to this and no need the if?

window.gtag?.("event", eventName, eventParameters);

src/analytics/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const TRACKING_EVENTS = {
```
utas-raymondng, line 1: use enum easier?

export enum TrackingEventName {
      CLICK: "click"
}

src/analytics/useAnalytics.ts
```diff
@@ -0,0 +1,8 @@
+import { useCallback } from "react";
```
utas-raymondng, line 1: Given the dependency is [], so you do not want to evaluate again, since you already define the function outside of the function, you can directly import the function instead of create the hook, which basically the same and less complicated

src/components/list/SuggestedCitationList.tsx
```diff
@@ -23,14 +25,16 @@ const SuggestedCitationList: React.FC<SuggestedCitationListProps> = ({
   mode,
 }) => {
   const { checkIfCopied, copyToClipboard } = useDetailPageContext();
+  const { track } = useAnalytics();
```
utas-raymondng, line 28: Use the function directly without the need to use hook

utas-raymondng (approved):
LGTM

PR #545, opened 2025-09-18
Feature/6940 add search url params analytics

utas-amberbian:


src/analytics/searchParamsEvent.ts
```diff
@@ -0,0 +1,57 @@
+import { AnalyticsEvent } from "./analyticsEvents";
+import { trackCustomEvent } from "./customEventTracker";
+
+/**
+ * Parameter mapping from URL shortcodes to meaningful names
+ */
+const URL_PARAM_MAPPING = new Map([
```
utas-raymondng, line 7: No need to create this as it is duplicate of info use the paramReverseLookup in the UrlUtils.

In fact your approach may result in too many place that needs to change and track, is it easy to track the parameter in the rest api call level like searchReducer.tsx and inside all the searchResult functions. Track it before the call to axios.get () ?

utas-raymondng (changes requested):
Suggest approach change

src/components/common/store/searchReducer.tsx
```diff
@@ -116,6 +117,13 @@ const searchResult = async (
     p.sortby = param.sortby;
   }
 
+  // Track analytics only for list searches to avoid duplicate events
```
utas-raymondng, line 120: Please move the if check into the trackSearchUrlParameter as this is something track wants to know but not others

src/analytics/searchParamsEvent.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * OUTPUT TO GA4:
+ * {
+ *   "search_text": "123",
+ *   "bbox": "104,-43,163,-8",
+ *   "has_co_data": "true"
+ * }
+ */
+export function trackSearchUrlParameters(
+  searchParams: SearchParameters & { signal?: AbortSignal }
```
utas-raymondng, line 22: may I know why we have signal?

utas-raymondng (changes requested):
Minor change request

src/analytics/searchParamsEvent.ts
```diff
@@ -0,0 +1,64 @@
...
+ * }
+ */
+export function trackSearchUrlParameters(searchParams: SearchParameters) {
+  // Skip list searches
+  if (searchParams.properties !== "id,centroid") {
+    return;
+  }
+
+  // Create unique key for this search
+  const searchKey = `${searchParams.text || ""}_${searchParams.sortby || ""}`;
```
utas-raymondng, line 28: Lets say if I do not enter any text, you only capture the first search event even I change the filter value? This happens because text is empty all the time and sortby can always the same then you will have searchKey === lastTracked.

Please can you add test case so that you can guard your logic and give other better understand of what you want to do.

utas-raymondng (changes requested):
Request change

src/analytics/searchParamsEvent.ts
utas-amberbian: Yeah, but the current implementation is intentionally focused on one-time search tracking per our requirements. Including filter changes would generate too much noise from map movements.
I'll add a test case to document this. Thanks

utas-amberbian: Oh, I see we should only remove the bbox second time change. Let me update that

utas-amberbian: Oh wait, this ticket just said "from url param", so my current one should be correct

utas-raymondng: If you want to do this, that means it is not the data we capture not correct, it is how we are going to use it. Do you think it should be the job of the report on how to interpret the data, not the source of generation? 

So the idea is, the source should generate info correct and no ambiguity , then the user can write what ever report to sum it up.

Obviously there may be a cost concern on data storage what we work on the source of generation but we know what we are doing and justify

utas-raymondng: That said you can generate a tag by indicate this is due to map movement or what

And a tag to tell it is due to user input

PR #546, opened 2025-09-18
✨ add page response time tracking for GA4

utas-amberbian:


src/analytics/pageResTimeEvent.ts
```diff
@@ -0,0 +1,42 @@
+import { AnalyticsEvent } from "./analyticsEvents";
+import { trackCustomEvent } from "./customEventTracker";
+
+export function trackPageResponseTime(): void {
+  window.addEventListener("load", () => {
```
utas-raymondng, line 5: remove the event listener after use else keep adding listener?

so const t = () => {       
   windows.removeEventListern(...t );
}

windows.addEventListener("load", t);

May be I do not have enough knowledge to comment but this is single page app so there is only one load?

utas-raymondng (changes requested):
question?

PR #547, opened 2025-09-18
Feature/7116 enable wfs downloading subsetting

NekoLyn:
uuid: 9efa25cd-4da4-47b5-9385-45e3cbd11705
image
image

### DIFF
--- a/src/components/common/store/searchReducer.tsx
+++ b/src/components/common/store/searchReducer.tsx
@@ -22,7 +22,7 @@ import { FeatureCollection, Point } from "geojson";
 import { mergeWithDefaults } from "../../../utils/ObjectUtils";
 import {
   DatasetDownloadRequest,
-  IDownloadCondition,
+  WFSDownloadRequest,
 } from "../../../pages/detail-page/context/DownloadDefinitions";
 import {
   getDateConditionFrom,
@@ -54,12 +54,6 @@ export type SearchControl = {
   pagesize?: number;
   searchafter?: Array<string>;
   score?: number;
-};
-
-export type WFSDownloadRequest = {
-  uuid: string;
-  layerName: string;
-  downloadConditions: IDownloadCondition[];
 };
 
 type OGCSearchParameters = {
@@ -332,22 +326,13 @@ const processWFSDownload = createAsyncThunk<
 
       if (!response.ok) {
         return thunkAPI.rejectWithValue(
-          createErrorResponse(
-            response.status,
-            `HTTP error! status: ${response.status}`,
-            "WFS download request failed"
-          )
+          createErrorResponse(response.status, "WFS download request failed")
         );
       }
 
       return response;
     } catch (error) {
-      if (error instanceof Error) {
-        return thunkAPI.rejectWithValue(
-          createErrorResponse(500, error.message, "WFS download request failed")
-        );
-      }
-      return thunkAPI.rejectWithValue(error);
+      errorHandling(thunkAPI);
     }
   }
 );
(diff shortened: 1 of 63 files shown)


src/hooks/useWFSDownload.tsx
```diff
@@ -333,12 +343,21 @@ const useWFSDownload = (onCallback?: () => void) => {
...
+
+        // Start fetch with abort controller
         const controller = new AbortController();
         abortControllerRef.current = controller;
-        const timeout = process.env.VITE_WFS_DOWNLOADING_TIMEOUT
-          ? Number(process.env.VITE_WFS_DOWNLOADING_TIMEOUT)
-          : CONNECTION_TIMEOUT_DEFAULT; // Frontend 20 minutes timeout by default
-        const timeoutId = setTimeout(() => controller.abort(), timeout);
 
         const response = await fetch(
```
utas-raymondng, line 362: I think it is better to push the ogc api call to the store as it is easier to find all these ogc call later, you can reference the fetchWithoutStore.. etc

utas-raymondng (changes requested):
Minor change requested

src/components/common/store/searchReducer.tsx
```diff
@@ -270,6 +285,68 @@ const processDatasetDownload = createAsyncThunk<
...
+  { rejectValue: ErrorResponse }
+>(
+  "download/downloadWFS",
+  async (request: WFSDownloadRequest, thunkAPI: any) => {
+    try {
+      // Extract download conditions
+      const dateRange = getDateConditionFrom(request.downloadConditions);
+      const multiPolygon = getMultiPolygonFrom(request.downloadConditions);
+      const format = getFormatFrom(request.downloadConditions); // Currently unused - CSV only for WFS
+
```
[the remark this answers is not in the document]
NekoLyn: I have made some change to use thunkAPI.signal for abort with Fetch and works
the reason I didn't use axios is: Axios in browsers uses XMLHttpRequest doesn't support responseType: 'stream'  (the browser log shows: The provided value 'stream' is not a valid enum value of type XMLHttpRequestResponseType)

why don't use Axios: Axios/XMLHttpRequest cannot stream in browsers
why use Fetch: Fetch natively supports streaming with ReadableStream

src/components/common/store/searchReducer.tsx
```diff
@@ -275,6 +284,55 @@ const processDatasetDownload = createAsyncThunk<
...
+          responseType: "stream",
+          headers: {
+            "Content-Type": "application/json",
+            Accept: "text/event-stream",
+            "Cache-Control": "no-cache",
+          },
+          timeout: WFS_DOWNLOAD_TIMEOUT, // 20 minutes for WFS downloads
+          signal: thunkAPI.signal,
+        })
+        .then((response) => response.data)
```
utas-raymondng, line 328: I think you should return promise directly

      return axios
        .post("/api/v1/ogc/processes/downloadWfs/execution", requestBody, {
          adapter: "fetch", // Use fetch adapter for streaming
          responseType: "stream",
          headers: {
            "Content-Type": "application/json",
            Accept: "text/event-stream",
            "Cache-Control": "no-cache",
          },
          timeout: WFS_DOWNLOAD_TIMEOUT, // 20 minutes for WFS downloads
          signal: thunkAPI.signal,
        })

the caller will be const v = dispatch(processWFSDownload...)


and then you can use v.abort() to cancel the call 

or v.then() to extract the value.

utas-raymondng (changes requested):
Return promise is better

utas-raymondng (approved):
LGTM

PR #556, opened 2025-09-24
Bugfix/7438 show more result button

utas-raymondng:
The reason this bug happens is because there are two place for the very similar query, this change unify them and add test case for checking


NekoLyn (approved):
LGTM

PR #567, opened 2025-09-30
Feature/7461 slider on wms

utas-raymondng:
Make geoserver layer time slider aware, so tile will target on the range


HavierD (approved):
LGTM

PR #590, opened 2025-10-17
use non-specified to instead default date

HavierD:


HavierD:
don't merge until nirali is happy

utas-raymondng (approved):
LGTM

PR #593, opened 2025-10-20
Bugfix/7618 related resources crash

utas-raymondng:
Some malform json in the title of related resources which cause the page to crash, need another fix to understand why value copped


HavierD (approved):
LGTM

PR #594, opened 2025-10-20
Bugfix/7549 add geoserver multi layers

NekoLyn:
frontend demo:
feching layers:
image
layers dropdown and layer rendering:
image
mobile view:
image


utas-raymondng (approved):
LGTM

PR #616, opened 2025-11-03
Fix bug on search get cancel accidentally

utas-raymondng:
The if(needNavigate) is calling before the dispatch(...), the navigate will cause the location.state update and accidentally trigger

````
  useEffect(() => {
    const handleNavigation = () => {
      const reduxContents = getSearchQueryResult(store.getState());
      if (
        location.state?.referer === pageReferer.SEARCH_PAGE_REFERER &&
        location.state?.requireSearch === false
      ) {
        // If the referer is SEARCH_PAGE_REFERER, it means the user is interacting with the search page
        // Meanwhile if the state requireSearch is false, it means the user is not required to do a search. This happens when user just change the layout
        // However the referer = SEARCH_PAGE_REFERER and requireSearch = false will be persist but redux will be cleared when user refresh the page, so we need to do a full search
        // Therefore we need to check if the redux content is empty or not to decide whether to do a full search or no search
        if (reduxContents.result.total > 0) {
          return;
        } else {
          doListSearch();
        }
      } else if (
        // If user navigate from DetailPage, as the redux store has results content already we just need to do a map search
        // However when user refresh the page, the redux will be cleared, we need to do a full search
        // Therefore we need to check if the redux content is empty or not to decide whether to do a full search or only map search
        location.state?.referer === pageReferer.DETAIL_PAGE_REFERER &&
        location.state?.requireSearch === false
      ) {
        if (reduxContents.result.total > 0) {
          doMapSearch()?.finally(() => {});
        } else {
          doListSearch();
        }
      } else {
        // In other cases, we need to do a full search
        // This including (but not limit): user pastes the url directly, navigate from landing page, change sort ...
        doListSearch();
      }
    };
    handleNavigation();

    return () => {
      cancelAllSearch();
    };
    // Must use location.state as the value inside can be the same reference, and hence will not trigger execution
    // the state object will be different each time.
  }, [cancelAllSearch, doListSearch, doMapSearch, location.state]);

````

and cause the cancelAllSearch() called. There is a small delay on this useEffect() being invoke and therefore sometime search get cancel sometimes not.

### DIFF
--- a/playwright/tests/search_page/test_map.py
+++ b/playwright/tests/search_page/test_map.py
@@ -385,7 +385,6 @@ def test_map_resets_to_default_after_landing_page(desktop_page: Page) -> None:
     assert are_value_equal(current_map_zoom, default_map_zoom)
     assert are_coordinates_equal(current_map_center, default_map_center)
 
-@pytest.mark.skip("untable")
 @pytest.mark.parametrize(
     'data_id, data_lng, data_lat',
     [
@@ -422,7 +421,6 @@ def test_map_card_popup_download_button_in_desktop(
     expect(search_page.main_map).to_be_visible()
 
 
-@pytest.mark.skip("unstable")
 @pytest.mark.parametrize(
     'data_id',
     [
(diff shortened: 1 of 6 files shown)


shaunahu (approved):
LGTM

PR #621, opened 2025-11-05
:bug: fix geoserver layer popup

NekoLyn:
example: 
http://localhost:5173/details/4402cb50-e20a-44ee-93e6-4728259250d2?tab=summary

bug 1:
click event is removed and not registered again on layer changes -> no click popup when change layers

bug 2:
not clean popup on fetch error -> some empty popup appearing on failed map_feature call

bug 3:
when draw rectangle, on the second click, the mode immediately changes from "draw_rectangle" to "simple_select" -> `isDrawModeRectangle(map)` returns false and show popup on the second click


utas-raymondng (approved):
LGTM

PR #622, opened 2025-11-05
Fix layer name

utas-raymondng:
The layer name is not always the title name, if the user have a valid URL contains layername use that one instead of title.


NekoLyn (approved):
LGTM

PR #630, opened 2025-11-07
Bugfix/7732 spatial extents point

utas-raymondng:
Bug fix and add spatial extents point

Test with this uuid => 0bd6ebaf-f978-45c9-b351-639e1b84f06c


NekoLyn (approved):
LGTM

PR #647, opened 2025-11-17
Feature/7356 remember map location

utas-raymondng:
Remember last map location on detail page across tab. Another ticket will be used to fix map layer selection across tab https://github.com/aodn/backlog/issues/7765


NekoLyn (approved):
LGTM

PR #666, opened 2025-12-03
Bugfix/7464 subsetting icon no show

utas-raymondng:
1. The time slider do not show because the checking should include "date" in additional to "dateTime"
2. The useEffect in the DownloadCard  to set the download service isn't right, it should not be the upper layer job. Do it in DownloadCard cause logical problem especially GeoServer layer may rest it.


NekoLyn (approved):
LGTM

PR #669, opened 2025-12-09
:sparkles: support old portal url pattern

NekoLyn:


NekoLyn:
> Is doing this easier?
> 
> ```
> const router = createBrowserRouter([
>   {
>     path: "/user/:id",          // :id is the param
>     element: <UserProfile />,
>   },
>   {
>     path: "/search",
>     element: <SearchResults />,
>     // query params via useSearchParams()
>   },
> ]);
> ```
> 
> That is route based on path and argument possible?


in old portal url, the uuid is not in the path and cannot do like
```
{
  path: "/search?uuid=:uuid",  // Invalid
  element: <DetailsPage />
}
```
but we can either add loader function https://reactrouter.com/6.30.2/route/loader
```
{
  path: "/search",
  loader: ({ request }) => {
    const url = new URL(request.url);
    const uuid = url.searchParams.get("uuid");
    return uuid ? redirect(`/details/${uuid}`) : null;
  },
  element: <SearchPage />
}
```
or component-based logic
```
{
  path: "/search", 
  element: <SearchRedirectComponent />  // Checks query params inside component
}
```

utas-raymondng (approved):
LGTM

PR #679, opened 2025-12-18
Faster search by sort and binary lookup

utas-raymondng:
Sort and binary lookup for hexbin timeslider to improve speed


Nirali-IMOS (approved):
LGTM

PR #695, opened 2026-01-12
add the playwright-local env to avoid confuse with "dev" run

utas-raymondng:
In playwright test, we move the map via API, however in the code we only consider user action as valid action , like mouse move is valid if it is user init, not api init.

That case some playwright test fail, before this change we use "dev" env to indicate it is a playwright test together with normal pc local run, this is confuse and make some before not as expected for local run. Hence introduce a playwright-local env


NekoLyn (approved):
LGTM

PR #710, opened 2026-01-21
Bugfix/7973 missing attribution world imagery

utas-raymondng:
The world imagery comes from ArcGIS and is not something from Mapbox so the default attribution from Mapbox do not include it.


NekoLyn (approved):
LGTM

PR #722, opened 2026-01-28
Bugfix/7773 fix zoom level detail page map

NekoLyn:
the map layer `FitToSpatialExtentsLayer` execute fit to bound on detail page loading when summary panel is hidden that cause problem:
1. the `map.cameraForBounds()` can't calculate correct zoom when the container is hidden (so the init zoom level is incorrect)
2. the `map.cameraForBounds()` can't calculate correct zoom when the container is changing from hidden to visible (need to wait for map resize finish)
3. the `useEffect` in `FitToSpatialExtentsLayer` is not triggered again when tab switching (so added one dependency that reflect tab changes)

new bug found related to `FitToSpatialExtentsLayer` : the click in spatial coverage map cannot make the main map navigate to the focus area:
`map.once(RENDER)` is not fired when focus area changed, it is a one-time listener for the next render event

### DIFF
--- a/playwright/tests/detail_page/maps/test_map_interactions.py
+++ b/playwright/tests/detail_page/maps/test_map_interactions.py
@@ -74,6 +74,7 @@ def test_selecting_date_range_adds_download_filter(
         '0015db7e-e684-7548-e053-08114f8cd4ad',
     ],
 )
+@pytest.mark.skip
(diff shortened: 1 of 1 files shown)


NekoLyn:
two playwright tests are skipped and need to be reviewed
ticket raise https://github.com/aodn/aodn-portal-v2/issues/724

PR #729, opened 2026-01-30
:bug: fix click spatial coverage map navigation

NekoLyn:
~~1 in Geojson layer, on layer click, query the clicked bbox and pass to FitToSpatilaExtentsLayer for fitting to~~
2 `map.once(MapEventEnum.RENDER, () => fitToBound(map, b[0]))` does not run for bbox changing once the map is rendered and idle, so for bbox update we just call `fitToBound` if the style is loaded

uuid: 48cf3cb9-caa9-4633-9baa-8bba3c4d904a

image

### DIFF
--- a/src/components/map/mapbox/layers/FitToSpatialExtentsLayer.tsx
+++ b/src/components/map/mapbox/layers/FitToSpatialExtentsLayer.tsx
@@ -30,9 +30,17 @@ const FitToSpatialExtentsLayer: FC<FitToSpatialExtentsLayerProps> = ({
         map.getBounds()?.getEast() !== bbox?.getEast() ||
         map.getBounds()?.getNorth() !== bbox?.getNorth()
       ) {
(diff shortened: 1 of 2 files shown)


src/components/map/mapbox/layers/GeojsonLayer.tsx
```diff
@@ -38,7 +39,7 @@ const BLUE_PIN_NAME = "blue_pin_name";
 
 const GeojsonLayer: FC<GeojsonLayerProps> = ({
   collection,
-  onLayerClick = (_: MapMouseEvent) => {},
+  onLayerClick = (_: LngLatBounds) => {},
```
utas-raymondng: Not good to supply empty handler, can you use undefined by default and then when call the handler use onLayerClick?.(bound)

utas-raymondng (changes requested):
Odd that need to skip test must be something not right

PR #740, opened 2026-02-13
Feature/7956 polygon selection design update

utas-amberbian:


utas-amberbian:
> Minor change request

Hi @utas-raymondng,

Test added.

PR #741, opened 2026-02-13
:sparkles: add parameters card to detail page sidebar (#6962)

utas-amberbian:


src/pages/detail-page/subpages/side-cards/ParametersCard.tsx
```diff
@@ -0,0 +1,111 @@
...
+  );
+
+  return (
+    <SideCardContainer title="Parameters">
+      <Stack direction="row" flexWrap="wrap" gap={1}>
+        {categoryLabels.map((label) => (
+          <Chip
+            key={label}
+            label={label}
+            size="small"
```
NekoLyn: could use `Link` here? so user could right click to open in new tab
or make it always open in a new tab so they won't lose previous page?

src/pages/detail-page/subpages/side-cards/ParametersCard.tsx
```diff
@@ -0,0 +1,111 @@
...
+          theme.concepts[0].title === PARAMETER_VOCAB_TITLE
+      )
+      .flatMap((theme) => theme.concepts)
+      .forEach((concept) => {
+        const category = conceptToCategoryMap.get(concept.id);
+        if (category) categories.add(category);
+      });
+
+    return [...categories].sort((a, b) => a.localeCompare(b));
+  }, [collection, conceptToCategoryMap]);
```
NekoLyn: the work looks heavy here: fetch all parameters and extract labels and filter by `collection.theme.concept.id`
besides, the vocabs are not always explicit in themes so for many records there is no way to find the vocabs in this way 
so I think it is better for frontend to get the `parameter_vocabs` from the collection directly
since we already generated parameter vocabs for each record during es indexing, so we just need to update OGC API converter to put this property in the collection 
@utas-raymondng what do you think?

NekoLyn: pr made for ogc api change https://github.com/aodn/ogcapi-java/pull/242 @utas-amberbian

src/pages/detail-page/subpages/side-cards/ParametersCard.tsx
utas-amberbian: Nope, it's a bit out of scope

PR #743, opened 2026-02-17
Feature/7978 add delivery mode tag

shaunahu:


src/pages/detail-page/subpages/HeaderSection.tsx
```diff
@@ -173,22 +175,43 @@ const renderSubTitle = (
         />
       </RoundCard>
     )}
-    {pace &&
-      pace.toLowerCase() !== "other" &&
-      !(pace.toLowerCase() === "completed" && status === Status.completed) && (
-        <RoundCard
+    {aiUpdateFrequency &&
+      aiUpdateFrequency.toLowerCase() !== "other" &&
+      !(aiUpdateFrequency.toLowerCase() === "completed") && (
```
NekoLyn: could create enum for "other"   "completed" etc

src/styles/constants.ts
```diff
@@ -128,6 +129,7 @@ const color = {
     light: "#979797",
     extraLight: "#dfdfdf",
     xxLight: "#f8f8f8",
+    warm: "#F6F0ED",
```
NekoLyn: we could use custom themes library. try to avoid add more constants here

src/components/result/ListResultCard.tsx
```diff
@@ -52,7 +52,64 @@ const ListResultCard: FC<ListResultCardProps> = ({
     const isSelectedDataset = uuid === selectedUuid;
     const thumbnail: string = findThumbnail();
     const scope = content.getScope();
+    const aiUpdateFrequency = content.getAiUpdateFrequency();
+    const hasAiUpdateFrequency =
+      aiUpdateFrequency?.toLowerCase() === "real-time" ||
+      aiUpdateFrequency?.toLowerCase() === "delayed";
```
NekoLyn: could create enums for `aiUpdateFrequency`

and could wrap them in useMemo to avoid unnecessary re-rendering

src/components/result/ListResultCard.tsx
```diff
@@ -169,50 +226,25 @@ const ListResultCard: FC<ListResultCardProps> = ({
...
-                        color: fontColor.black.dark,
-                        lineHeight: lineHeight.heading,
-                      }}
-                    />
-                  </Box>
-                )}
+                {!shouldHideTags && hasDocumentTag && renderTagChip("Document")}
+                {!shouldHideTags &&
+                  hasAiUpdateFrequency &&
+                  renderTagChip(aiUpdateFrequency || "other")}
```
NekoLyn: do you want to double check with Nat, since previously they don't want the "other" label, said this is confused for users as they don't know the context

PR #744, opened 2026-02-18
:truck: refactor: rename subpages folders

utas-amberbian:


utas-raymondng:
Also it impacted many files which I feel risky to do it in one go

PR #757, opened 2026-03-02
🐛 fix draw mode not switching between bbox and polygon

utas-amberbian:


src/components/map/mapbox/controls/menu/DrawRect.tsx
```diff
@@ -68,7 +68,8 @@ const DrawRect: React.FC<DrawControlProps> = ({
       mapDraw.changeMode(DRAW_RECTANGLE_MODE);
       setShowTooltip(true);
     } else {
-      // If already in draw mode, just show tooltip again
+      // If already in draw mode, switch to rectangle and show tooltip
```
utas-raymondng: Looks strange logic, so effectively the else if no use???


else if (!isDrawingMode) {
      // If not in draw mode, activate draw mode and show tooltip
      mapDraw.changeMode(DRAW_RECTANGLE_MODE);
      setShowTooltip(true);
    } else {
      // If already in draw mode, switch to rectangle and show tooltip
      mapDraw.changeMode(DRAW_RECTANGLE_MODE);
      setShowTooltip(true);
    }

PR #760, opened 2026-03-03
:bug: fix download size format and wrapping

NekoLyn:
**change format to include KB, MB and GB, so in most cases text won't get truncated**
Image

**wrap text to two lines if it's too long in special screen size**
Image


utas-raymondng (approved):
LGTM

PR #783, opened 2026-04-09
Fix strange behavior in input handling

utas-raymondng:
1. If you type something in input then keep pressing backspace, there will be a racing condition that words will reappear
2. Select word ignore in search if you select words in the drop down and immediate press return


NekoLyn (approved):
LGTM

PR #784, opened 2026-04-10
Bugfix/7856 inactive when lost focus

utas-raymondng:
Slightly improve the behavior of searchbar, keep expand when user click the x button but not expand if the text field is empty and lost focus


NekoLyn (approved):
LGTM

PR #787, opened 2026-04-13
add http request errors event for ga4

utas-amberbian:


utas-raymondng (changes requested):
Any test verify execution as expected?

utas-raymondng (approved):
LGTM, will there update on GA report?

utas-amberbian:
> LGTM, will there update on GA report?

Not yet — this is just for testing at the moment. The plan is to think through some use cases first, then update BigQuery and add a report after that.

PR #788, opened 2026-04-13
Task/8180 degrade page optimization

HavierD:
don't use dedicated path `/degraded` so that user can refresh to back / system can auto refresh every 30s.


utas-raymondng (approved):
LGTM

PR #789, opened 2026-04-14
Bugfix/7524 misalign in firefox and header image/text scaling

utas-raymondng:
Fix the following 
1. The hamburger is a bit move to right when using firefox (no space on right) but no issue with edge
2. The logo and text update to fit the mobile design
3. hamburger deign update

## Before
image

## After
image


utas-amberbian (approved):
LGTM

PR #790, opened 2026-04-15
Bugfix/7524 banner text mobile misalign

utas-raymondng:
Refactor theme to make it responsive avoid use of useBreakpoint() everywhere, the breakpoint is same as what we have now.

## Before
image

## After
image


utas-amberbian (approved):
LGTM

PR #791, opened 2026-04-16
Bugfix/7524 input with suggester layout fix

utas-raymondng:
1. Fix issue on drop down position

before

https://github.com/user-attachments/assets/74c47958-159c-42b1-bceb-73d029aef6a9

after fix the dropdown do not block the button.

2. Add the search button on the highlight item in the suggest list

https://github.com/user-attachments/assets/1e46823b-88f9-42ad-bdbe-e733302ad752

3. If too many words, then move the button to next row

https://github.com/user-attachments/assets/721f3b24-c700-4d1a-b8f5-be96e1e06711

### DIFF
--- a/src/components/search/InputWithSuggester.tsx
+++ b/src/components/search/InputWithSuggester.tsx
@@ -17,6 +17,7 @@ import {
   Popper,
   TextField,
 } from "@mui/material";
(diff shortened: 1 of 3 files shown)


utas-amberbian (approved):
LGTM

PR #792, opened 2026-04-17
Bugfix/8321 daterangefilter match mobile design

utas-raymondng:
Fix date time filter drop down, there are some issue with the design and need to address by designer


utas-amberbian:
Hi @utas-raymondng,
A couple of small comments:
1. Colour consistency
Could the colour be adjusted to align with the design spec?
Screenshot from 2026-04-20 16-30-28
2. Possible bug
There also seems to be a small bug here — could you take a look?
Screenshot from 2026-04-20 16-28-04
Thanks!

utas-raymondng:
Fixed, I have a quick chat with @quye04 and he is ok to have the calendar icon missing here in mobile version

image

I think it is too much to hack into the component just to make the icon visible and maintain the click behavior. By default it detect the screen and if it is a mobile, it render differently which use the mobile version. This will by default disable the icon, if you want to add it back then you will break the click action

src/components/filter/DateRangeFilter.tsx
```diff
@@ -401,56 +439,77 @@ const DateRangeFilter: FC<DateRangeFilterProps> = memo(
...
+                      slots={{
+                        openPickerIcon: CalendarIcon,
+                      }}
+                      slotProps={{
+                        ...DEFAULT_DATE_PICKER_SLOT,
+                        openPickerIcon: {
+                          color: theme.palette.grey600,
+                          width: 22,
+                          height: 22,
+                        } as any,
```
utas-amberbian: would satisfies SVGProps<SVGSVGElement> or other types work here instead of as any

src/components/details/DateRangeSlider.tsx
```diff
@@ -14,67 +14,75 @@ import { LocalizationProvider } from "@mui/x-date-pickers/LocalizationProvider";
...
+    ".MuiPickersCalendarHeader-root": {
+      marginTop: "8px",
+    },
+    ".MuiPickersYear-yearButton": {
+      color: theme.palette.text2,
+      padding: 0,
+    },
+    ".MuiPickersYear-yearButton.Mui-selected, .MuiPickersYear-yearButton.Mui-selected:hover, .MuiPickersYear-yearButton.Mui-selected:focus":
+      {
+        color: "#fff !important",
```
utas-amberbian: is there a specificity issue that needed !important here? there are a lot of ! important, which will be a tech debt.

src/components/filter/DateRangeFilter.tsx
```diff
@@ -401,56 +439,77 @@ const DateRangeFilter: FC<DateRangeFilterProps> = memo(
...
```
utas-amberbian: also here thanks

utas-amberbian (approved):
LGTM

PR #793, opened 2026-04-21
Playwright: fix skipped test

AR-RIFAT:
1. Fixed skipped test
2. Added layer select loading to avoid failure in detail page tests
3. Added function for menu item clicks to avoid MUI popover issue


utas-raymondng (approved):
LGTM

PR #795, opened 2026-04-24
Feature/8322 refactor before change

utas-raymondng:
The DrawRect depends on argument that is not generic which  makes it hard to reuse elsewhere


utas-amberbian (approved):
LGTM

PR #798, opened 2026-04-30
Feature/8086 search page new breakpoint

NekoLyn:
#### Main changes:
- change laptop breakpoint from 1280px to 1024px
- update layout in search page: result card size, sort/layout button display, full list view display, gaps and paddings in laptop/tablet/mobile

---

#### Other UI affected fix:
- landing page padding in laptop/tablet/mobile and other affected UI like footer
- detail page padding in laptop/tablet/mobile and other affected UI like goback button

---

#### Pending to implement:
1. result card button group: design still in progress and need discuss and feedback
image
<br>
2. search result simple filter bar: need to consider the share button and re-design
image

---

#### Note:
1. as discussed with Steve, some small size adjustment happen in search page result cards width and map width since the scroll bar is better to stay together with the scroll-able section.
image
<br>

2. the list result card layout is the "top-bottom" pattern while the design is "left-right" like pattern - keep the current "top-bottom" pattern as design missing the case when there is no thumbnail which may cause some UI problem.

3. please feel free to test all the UI components and functions that may affected in difference screen size and let me know if there is an issue, thanks!

---

#### FE demo:
image
image
image


utas-amberbian (approved):
LGTM

PR #808, opened 2026-05-05
Feature/8322 update location filter

utas-raymondng:
1. Implement the design proposed.
2. New property introduced staticArea to hold area of selection, the polygon will hold the polygon that is draw manually.
3. Some test update to reflect the change, but the url test is now skip and need playwright test fix as new property introduced. https://github.com/aodn/aodn-portal-v2/issues/814


utas-amberbian (approved):
LGTM

PR #815, opened 2026-05-07
Feature/7526 update detail page tab panel icons

NekoLyn:
main change:
use icon for tabs in mobile screen
image


src/components/common/tab/StyledTab.tsx
```diff
@@ -6,16 +6,20 @@ interface StyledTabProps extends TabProps {
   showBadge?: boolean;
   badgePosition?: Position;
   badgeColor?: string;
+  isMobileText?: boolean;
```
utas-amberbian, line 9: I think there are some mobile size [here 
](https://github.com/aodn/aodn-portal-v2/blob/main/src/styles/themeRC8.ts)

utas-amberbian:
For our SVG icons, should we put them in components/icon or assets/icon? Want to make sure we're consistent across the codebase.

src/components/common/tab/StyledTab.tsx
NekoLyn: Thanks for point out that. I think add tab font size with breakpoint and use the `variant` will introduce more complexity, since the `isMobileText` come with other changes not just font size:
```
          ...(isMobileText && {
            ...portalTheme.typography.body2Regular,
            fontSize: "14.4px",
            height: "39px",
            minHeight: "39px",
            py: "9.5px",
            px: "20px",
          }),
```
so I prefer to keep it.

NekoLyn:
> For our SVG icons, should we put them in components/icon or assets/icon? Want to make sure we're consistent across the codebase.

Thanks. I noticed there are two places for icons. I think both of them contains reusable React component icons, good put in one place. I created a ticket for refactor folder structure for icons and other static resources https://github.com/aodn/backlog/issues/8434

utas-amberbian (approved):
LGTM

PR #816, opened 2026-05-07
Fix filter missing time cause ogcapi error

utas-raymondng:
This fix is make sure the time is pass on the query, before change only date part is set and time part all zero which cause a problem with end data if you set the start date = end date.

After the fix the time part will pass together align with the end date value which is usually 23:59:59


utas-amberbian (approved):
LGTM

PR #818, opened 2026-05-08
do not render hexbin for AMSA

shaunahu:


shaunahu:
closed as we decide to block from ogc-api side

PR #819, opened 2026-05-08
Feature/7526 map as separate tab in mobile

NekoLyn:
main changes:
1. add separate map tab in mobile screen
2. make summary show all by default in mobile screen
image
image


utas-amberbian (approved):
LGTM

PR #825, opened 2026-05-15
:bug: fix CO subset date-range slider

NekoLyn:
main change: 
(left) time slider: reflects Geoserver layer temporal (comes from ncWMS layer info)
(right) date range slider: reflects download subset support temporal (comes from CO feature summary)
image


utas-amberbian (approved):
LGTM

PR #826, opened 2026-05-18
Feature/8376 filter select result update

utas-raymondng:
1. Add chips to allow remove individual filters
2. Remove the cross and rotate button and add the clear all button
3. The alignment is not 100% in search screen, there is a new update on design so will fix it in future ticket https://github.com/aodn/backlog/issues/7855#issuecomment-3781857661


NekoLyn (approved):
LGTM

PR #827, opened 2026-05-21
Bugfix/8376 word missing in chips

utas-raymondng:
1. The bug happens because when parse the URL param, it think the id is a number but the lookup from the chip assume it is string, the cause the compare failed and show empty.
2. It now show hide the divider by checking filter is empty or not.
3. Update chip delete action so it refresh search and update URL


NekoLyn (approved):
LGTM

PR #830, opened 2026-05-22
:bug: fix navigation bar observers

NekoLyn:
Use useEffect (not useLayoutEffect): 
- refs for the sibling DetailSubtabBtn list are attached AFTER VerticalIndicator's commit-time effects
- so a useLayoutEffect here sees `menuRefs[i].current === null` and never wires up the two observers `ResizeObserver`/`IntersectionObserver`
- useEffect runs after the full tree commits and paint, when every ref is populated.


utas-raymondng (approved):
LGTM

PR #833, opened 2026-05-27
:bug: fix filter chips redirect and style

NekoLyn:


utas-raymondng (approved):
LGTM, if you can move the suggester down a bit on search page if the filter exist will be perfect

PR #839, opened 2026-06-05
add ai vocabs for parameter and platform filter

shaunahu:
Skip playwright test `test_search_api_request_urls_after_map_state_change` in `test_request_urls.py` because the search reducer has been updated and the related playwright test should be updated accordingly. (Issue ticket raised in https://github.com/aodn/aodn-portal-v2/issues/842)

**Changes**
- Parameter filters now query both `parameter_vocabs` and `ai_parameter_vocabs`.
- Platform filters now query both `platform_vocabs` and `ai_platform_vocabs`.
- Updated the search reducer unit test to expect AI parameter vocabs in generated filters.
- Temporarily skipped the Playwright API URL test for map state changes because its assertions need to be updated for the new search filter format.

FE demo:
image


utas-raymondng (approved):
LGTM

PR #847, opened 2026-06-25
revert: remove acronym full names from search autocomplete

utas-amberbian:


utas-raymondng (changes requested):
Question is there a need to change UI if there is no input suggestion change?

utas-raymondng (approved):
LGTM - just to rollback changes

PR #854, opened 2026-07-09
Feature/8070 add document filter

shaunahu:
Adds a new Exclude Documents toggle to the Data Settings filter tab. When enabled, the search request appends the CQL clause NOT (scope='document'), removing document-scope records from results.
image

The Gridded Datasets and Time Series Datasets topic cards on the landing page now apply excludeDocument = true and hasCOData = true automatically, so those shortcuts return only downloadable, non-document datasets.
image


NekoLyn (approved):
LGTM

PR #855, opened 2026-07-09
fix: center detail maps for antimeridian and near-global bbox

utas-amberbian:
Screenshot 2026-07-17 at 12 07 38
Screenshot 2026-07-17 at 12 08 09
Screenshot 2026-07-17 at 12 08 41
Screenshot 2026-07-17 at 12 09 18
Screenshot 2026-07-17 at 12 09 43
Screenshot 2026-07-17 at 12 09 43
Screenshot 2026-07-17 at 12 10 10
Screenshot 2026-07-17 at 12 10 48
Screenshot 2026-07-17 at 12 11 12

### DIFF
--- a/src/utils/MapUtils.tsx
+++ b/src/utils/MapUtils.tsx
@@ -1,5 +1,6 @@
 import { Root } from "react-dom/client";
 import {
+  LngLat,
   LngLatBoundsLike,
   Map as MapboxMap,
   MercatorCoordinate,
@@ -75,14 +76,21 @@ const half = worldSize / 2;
 const ensureEastGreaterThanWest = (west: number, east: number): number =>
   east < west ? east + 360 : east;
 
-const AUSTRALIA_CENTER: [number, number] = [
+const AUSTRALIA_CENTER_LNG =
   (MapDefaultConfig.BBOX_ENDPOINTS.WEST_LON +
     MapDefaultConfig.BBOX_ENDPOINTS.EAST_LON) /
-    2,
-  (MapDefaultConfig.BBOX_ENDPOINTS.SOUTH_LAT +
-    MapDefaultConfig.BBOX_ENDPOINTS.NORTH_LAT) /
-    2,
-];
+  2;
+
+// Only recentre when the bbox actually reaches Australia's longitudes,
+// otherwise data such as an Atlantic-hemisphere bbox would end up off-screen
+const overlapsAustraliaLng = (west: number, east: number): boolean => {
+  const { WEST_LON, EAST_LON } = MapDefaultConfig.BBOX_ENDPOINTS;
+  return (
+    Math.max(west, WEST_LON) <= Math.min(east, EAST_LON) ||
+    // for bboxes unwrapped past 180, Australia repeats at +360
+    Math.max(west, WEST_LON + 360) <= Math.min(east, EAST_LON + 360)
+  );
+};
 
 // A bbox this wide (or tall — catches degenerate metadata like
 // [180, -71, 180, 63]) fits at world-level zoom anyway, so the view is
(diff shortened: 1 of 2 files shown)


NekoLyn:
Hi, from the code seems like if the wide of the bbox is over `NEAR_GLOBAL_LNG_SPAN = 350`  it will fit to the default Australia bound. I am not sure about the requirement of this ticket, but see some potential issue:
- if the data are not around Australia, fit to Australia may cause seeing less data
- fit to the default Australia bound will change the original zoom level, missing more data on map

utas-amberbian:
> Hi, from the code seems like if the wide of the bbox is over `NEAR_GLOBAL_LNG_SPAN = 350` it will fit to the default Australia bound. I am not sure about the requirement of this ticket, but see some potential issue:
> 
> * if the data are not around Australia, fit to Australia may cause seeing less data
> * fit to the default Australia bound will change the original zoom level, missing more data on map

Could you share a few concrete examples of affected collections? That would make it easier to understand the concern. Thanks

NekoLyn:
> Could you share a few concrete examples of affected collections? That would make it easier to understand the concern. Thanks

**Hi, I only check the code itself, so the concern is raised from the code logic, so might not be able to find uuid to approve.**
- Didn't find record to meet the first point, guess most of our records are around Australia
- And for the second point, taking the record 997c2e15-b345-438f-afac-49a4ac19be38 listed in the ticket as an example:
  - before change: zoom level is small, can see data over the world
image

  - after change: zoom level is high, only see data around Australia
image

**Another two issue:**
- Using `NEAR_GLOBAL_LNG_SPAN = 350` the magic number would fail to center Australia if the width is 349
  - example 95d6314c-cfc7-40ae-b439-85f14541db71 the width is 230
- Found a record ca15915b-38c5-4e95-8d25-9cc42c4dd485 with strange extent which fail to pass exam width logic:
```
 "bbox": [
                [
                    180.0,
                    -70.0,
                    180.0,
                    50.0
                ],
                [
                    -180.0,
                    -70.0,
                    180.0,
                    50.0
                ]
            ],
```
the overall bounding box should be [-180, xx, 180, xx] - this might be the backend conversion error, need to double check
```
 [
   -180.0,
    -70.0,
    180.0,
    50.0
   ],
 ```

**Comment by NekoLyn** [comment]
If you don't mind, let's hold this ticket and discuss with Alex/Raymond and reach a conclusion about what is the expected final image.
Thanks.

File: src/utils/MapUtils.tsx [hunk 1]
```diff
@@ -70,6 +70,34 @@ const BBOX_COORDINATES_COUNT = 4;
...
+const WORLD_SCALE_LNG_SPAN = 180;
+const WORLD_SCALE_LAT_SPAN = 90;
+
+const isWorldScale = (
+  west: number,
+  south: number,
+  east: number,
+  north: number
+): boolean =>
+  east - west >= WORLD_SCALE_LNG_SPAN || north - south >= WORLD_SCALE_LAT_SPAN;
```
NekoLyn, line 107: I think it's span check only which doesn't involve the data location check which may cause the recenter AU occurs for data off AU ?

src/utils/MapUtils.tsx
```diff
@@ -101,23 +129,28 @@ export const fitToBound = (
...
 
       // Use flyTo for more control over the viewport
       map.flyTo({
-        center: center,
-        zoom: zoom,
+        center: isWorldScale(west, south, east, north)
+          ? AUSTRALIA_CENTER
+          : camera.center,
+        zoom: camera.zoom,
         animate: animate,
```
NekoLyn, line 172: fly to `AUSTRALIA_CENTER` will move the view vertically - so could keep the original lat just change the lon

NekoLyn (approved):
Hi, I have a couple of concerns about this change — flagging them as optional since all the examples Alex provided are fixed by the current logic:

1. **`isWorldScale` is a span-only check** — it never verifies the bbox actually covers Australia. A bbox like `[-180, -60, 0, 60]` (the Atlantic hemisphere, which doesn't touch Australia's lng range 104–163 at all) still gets recentred on Australia, so the actual data ends up at the screen edges or off-screen.

2. **`flyTo` with `AUSTRALIA_CENTER` also moves the view vertically** — it overrides the latitude that `cameraForBounds` computed. For a dataset concentrated at high latitudes (e.g. mainly around Antarctica), forcing the centre to lat -25.5 can push the data partially or fully out of view.

3. They are not fully fix what mentioned in the previous discussion:
>Map centering rules
>Data outside AU → center on data's own extent
>Data includes AU → center longitude on Australia
>Default extent unchanged

Since Alex's examples don't hit these edge cases, it's up to you whether to keep refining or leave it as is.

PR #857, opened 2026-07-13
feat(seo): index each route via per-route canonical URL

utas-amberbian:
branch renamed
https://github.com/aodn/aodn-portal-v2/pull/853


utas-raymondng (approved):
LGTM

PR #861, opened 2026-07-16
fix dataset types and map layer select

NekoLyn:
example record: 4ac6bf81-cd37-4611-8da8-4d5ae5e2bda3
image


utas-raymondng (approved):
LGTM

PR #864, opened 2026-07-21
Fix bug where it makes data not match between month and daily grouping

utas-raymondng:
Make UI able to process pmtile that is group by month or group by date


NekoLyn (approved):
LGTM

PR #871, opened 2026-07-22
Bugfix/8751 speed up pmtiles

utas-raymondng:
1. Fix speed issue
2. Use .metadata file to know the data is by date or by month
3. timeslider use min/max value from .metadata
4. Remove hexbin layer, more clean up is needed, just remove it from menu and map layer, but Hexbin code is still in repo


NekoLyn (approved):
LGTM

PR #884, opened 2026-07-28
Bugfix/8736 fix dataset status mapping

shaunahu:
FE demo:
1. status = "complete" (example uuid: `45d0c70c-ee4d-4727-8d86-daacb2d29156`)

Image

2. status = "onGoing | historicalArchive" (example uuid: `ddfe08f0-e4bc-11dc-9b6b-00188b4c0af8`):

Image

### DIFF
--- a/src/components/result/ResultCardButtonGroup.tsx
+++ b/src/components/result/ResultCardButtonGroup.tsx
@@ -29,10 +29,12 @@ interface ButtonContainerProps {
   sx?: SxProps;
 }
 
-enum Status {
-  Ongoing = "ongoing",
-  Completed = "completed",
-}
+// Lowercased status values accepted for each button, as the records use several
+// spellings for the same status and may combine codes into one string
+const Status = {
+  Ongoing: ["ongoing", "ongoing | historicalarchive"],
+  Completed: ["completed", "complete"],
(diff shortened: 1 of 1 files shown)


src/components/result/ResultCardButtonGroup.tsx
```diff
@@ -40,7 +40,7 @@ const renderStatusButton = (
   resultCardButtonConfig?: ResultCardButtonConfig
 ) => {
   const status = content?.getStatus()?.toLowerCase().trim();
-  if (status === Status.Completed) {
+  if (status === Status.Completed || status === "complete") {
```
NekoLyn: could you add "complete" to Status.Completed array so here you can use something like .include() to check if any matching

src/components/result/ResultCardButtonGroup.tsx
```diff
@@ -50,7 +50,7 @@ const renderStatusButton = (
       />
     );
   }
-  if (status === Status.Ongoing) {
+  if (status?.startsWith(Status.Ongoing)) {
```
NekoLyn: same like above

NekoLyn (approved):
LGTM

PR #885, opened 2026-07-28
Bugfix/8751 time slider boundary

utas-raymondng:
1. Fix issue with time slider where start and end is the same
2. Do not show popup when the hexbin have zero count, do not show the hexbin too
3. Tune constant to make the draw looks faster
4. Refactor to reduce the size of the file


NekoLyn (approved):
LGTM

PR #891, opened 2026-07-31
Bugfix/8868 dataset without time

utas-raymondng:
1. The pmtiles now contains dataset without time, the timeslider will be disable or invisible for those dataset.
2. pmtile marker for date aggregation file is now dYYYYMMDD, for monthly aggretation is mYYYYMM.


NekoLyn (approved):
LGTM

PR #896, opened 2026-08-04
Bugfix/8906 hide time subsetting

utas-raymondng:
1. Make sure if the layer do not support time slider, the download subsetting is hidden. Example e41efa35-03f0-4dea-be98-2a69e46b510b


NekoLyn (approved):
LGTM

PR #897, opened 2026-08-04
Feature/8276 update material UI

utas-raymondng:
1. Update to use materialui v7
2. May have some small discrepancy , however in order no to block everyone it is better to fix it with other PR


NekoLyn (approved):
LGTM

